PR #26, opened 2020-05-21
Explicitly sets IsolationLevel

rbans96:
## Description
This is to explicitly set IsolationLevel.
Default IsolationLevel is ReadCommitted.
If for any transaction we need to set different IsolationLevel, we can pass that in the SqlConnectionWrapper.
E.g. We need to set IsolationLevel.Snapshot in SqlServerSearchService

## Related issues
Addresses #70783

## Testing
Existing tests continue to pass.


rbans96:
It is no longer required as mentioned in the #[AB70783](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=70783)

PR #67, opened 2020-07-24
Adds IIndex<TService> to resolve a service with metadata

brendankowitz:
## Description
Adds the ability to register services with Metadata, then resolve this information with IIndex

For example, register all services implementing IComponent with the DisplayName attribute value as metadata
```csharp
            _collection.TypesInSameAssemblyAs<IComponent>()
                .AssignableTo<IComponent>()
                .Scoped()
                .WithMetadata(type => type.GetAttribute<DisplayNameAttribute>()?.DisplayName)
                .AsSelf()
                .AsImplementedInterfaces();
```

This can then be resolved later and the instance can be fetched with this value:

```csharp
            var index = container.GetService<IIndex<IComponent>>();

            IComponent componentAObj = index["Component A"];
            IComponent componentBObj = index["Component B"];
```

## Testing
Adds unit tests


brendankowitz:
Its coming it .NET 8 😄 https://twitter.com/davidfowl/status/1683956501570879489

PR #100, opened 2020-09-23
Fixes apply full schema snapshot

rbans96:
## Description
This change enables to apply x.sql even without the force option if the current schema version is null, where x is the maximum supported schema version.

## Related issues
Addresses [#75643](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=75643)

## Testing
Manually tested by running the schema migration tool.

- Started the server on schema version null
- Hit the command below command to apply latest supported version
 Microsoft.Health.SchemaManager apply --latest -cs "server=(local);Initial Catalog=SHARED_HEALTHCARE;Integrated Security=true" -s https://localhost:63637/

image

### DIFF
--- a/tools/SchemaManager/SchemaDataStore.cs
+++ b/tools/SchemaManager/SchemaDataStore.cs
@@ -30,11 +30,7 @@ namespace SchemaManager
 
                     serverConnection.BeginTransaction();
 
-                    // Since version 3 is just the base schema which must already be executed at the start of apply command and can be skipped
-                    if (version != 3)
-                    {
-                        server.ConnectionContext.ExecuteNonQuery(script);
-                    }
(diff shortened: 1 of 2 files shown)


tools/SchemaManager/SchemaDataStore.cs
```diff
@@ -30,11 +30,7 @@ public static void ExecuteScriptAndCompleteSchemaVersion(string connectionString
 
                     serverConnection.BeginTransaction();
 
-                    // Since version 3 is just the base schema which must already be executed at the start of apply command and can be skipped
```
rbans96, line 33: This is not related to the fix but identified this version 3 specific to fhir-server should be removed. As this is the common tool across dicom, fhir and any future repos.

brendankowitz (approved):

PR #101, opened 2020-09-25
Auditing executed messages in the middleware.

v-shaaal:
## Description
Auditing executed messages in the middleware. Before, executed messages were getting logged in two different places: Audit Middleware and Audit Logging Filter Attribute. Now, Audit Logging Filter Attribute will only be responsible to log the Executing messages. Executed messages will only be logged through Audit Middleware.

## Related issues
[AB#76278](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76278).

## Testing
- Tests added to verify that middleware is logging in the cases of success, in cases of different exceptions, and is also logging on auth failure.
- All the already existing tests are also passing.


src/Microsoft.Health.Api/Features/Audit/AuditEgressMiddleware.cs
```diff
@@ -43,14 +43,7 @@ public async Task Invoke(HttpContext context)
...
                 var statusCode = (HttpStatusCode)context.Response.StatusCode;
-
-                // Since authorization filters runs first before any other filters, if the authorization fails,
-                // the AuditLoggingFilterAttribute, which is where the audit logging would normally happen, will not be executed.
-                // This middleware will log any Unauthorized request if it hasn't been logged yet.
-                if (statusCode == HttpStatusCode.Unauthorized)
-                {
-                    _auditHelper.LogExecuted(context, _claimsExtractor);
-                }
+                _auditHelper.LogExecuted(context, _claimsExtractor);
```
brendankowitz, line 46: I'm not sure what the net effect is, but in thinking through it, we may have mismatched AuditExecuted events. Have you testing what happens with the custom endpoints like `/health/check` and [SchemaController](https://github.com/microsoft/healthcare-shared-components/blob/master/src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs)?

v-shaaal: What is meant by mismatched AuditExecuted events?
I did test for /health/check and it is working as before i.e. not getting logged. According to AuditHelper (in FHIR and DICOM), audit entries are only mapped for actions with assigned Audit Event Type. Since /health/check does not have an assigned Audit Event Type, it is not logged. Moving LogExecuted to AuditEgressMiddleware has not changed the existing functionality. But I would like to test specific scenarios if you could guide me through them.

brendankowitz: Sorry, I meant the calls to IAuditHelper  => LogExecuting / LogExecuted. Ok cool, just wanted to ensure we covered these utility endpoints.

v-shaaal: Yep, I made sure those are getting called as before :)

v-shaaal: Actually I did not know about the concept of Batch in FHIR. This won't work because of how batch requests are handled. This implementation would end up logging executed just once for the entire batch instead of logging for each sub-request within the batch. I will close this PR and put something together such that FHIR is not affected.

v-shaaal:
I did not know about the concept of Batch in FHIR. This won't work because of how batch requests are handled. This implementation would end up logging executed just once for the entire batch instead of logging for each sub-request within the batch. I will close this PR and put something together such that FHIR is not affected.

PR #102, opened 2020-09-28
Managed Identity for Sql Server

v-shaaal:
## Description
Initial changes for managed identity code for Sql server. Introduced a change in the configuration file where UseManagedIdentity can be set (this is set to false by default). This value is used further to decide the form of login. Also, all the connections to Sql server now go through SqlConnectionHelper which fetches and sets the access token if managed identity is to be used.

## Related issues
[AB#76113](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76113).

## Testing
All the existing tests are passing.


src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -38,7 +38,7 @@ public class SqlConnectionWrapper : IDisposable
             }
             else
             {
-                SqlConnection = new SqlConnection(configuration.ConnectionString);
+                SqlConnection = SqlConnectionHelper.GetSqlConnection(configuration.ConnectionString, configuration.UseManagedIdentity);
```
jackliums, line 41: This won't work in the managed service because how access token is obtained is different in SF vs AKS vs here.

What we should do is have an interface that allows the initialization of the SqlConnection and then it can be overwritten by the IoC container for each specific service.

src/Microsoft.Health.SqlServer/SqlConnectionHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+        public static SqlConnection GetSqlConnection(string connectionString, bool useManagedIdentity)
+        {
+            EnsureArg.IsNotNullOrEmpty(connectionString);
+
+            SqlConnection sqlConnection = new SqlConnection(connectionString);
+
+            if (useManagedIdentity)
+            {
+                var credential = new DefaultAzureCredential();
+                var token = new Microsoft.Azure.Services.AppAuthentication.AzureServiceTokenProvider().GetAccessTokenAsync("https://database.windows.net/").Result;
```
jackliums, line 31: This is async called. It should not be blocking.

jackliums: Also, you don't want to create a new instance every time because the underlying AzureServiceTokenProvider caches the token and refreshes it for you so you would want to reuse it. otherwise, it will reauthenticate every time.

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
v-shaaal: Addressed in the new iteration.

src/Microsoft.Health.SqlServer/Configs/SqlServerConnectionType.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.SqlServer.Configs
+{
+    public enum SqlServerConnectionType
+    {
+        ManagedIdentity,
+        WindowsIntegratedAuth,
+        Default,
```
smithago, line 12: What is the default connection type? Can we give it a name like SqlUserNameAndPassword?

v-shaaal: Yeah, I was rather confused with the name so left it Default. I will change it to SqlUserNameAndPassword as it makes more sense.

src/Microsoft.Health.SqlServer/ISqlConnectionFactory.cs
```diff
@@ -0,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Data.SqlClient;
+
+namespace Microsoft.Health.SqlServer
+{
+    public interface ISqlConnectionFactory
```
smithago, line 10: Factory class is not the iterface, it is ISqlConnection or IHealthSqlConnection. 
with a method to GetSqlConnection(). Then you would have implementations like ManagedIdentitySqlConnection: ISqlConnection,  SqlUserNamePasswordSqlConnection: ISqlConnection. And then you would have a SqlConnectionFactory that would return the right implementation of ISqlConnection

src/Microsoft.Health.SqlServer/Microsoft.Health.SqlServer.csproj
```diff
@@ -14,9 +14,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.2.3" />
```
smithago, line 17: Where is this reference used?

src/Microsoft.Health.SqlServer/DefaultSqlConnectionFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+using EnsureThat;
+using Microsoft.Health.SqlServer.Configs;
+
+namespace Microsoft.Health.SqlServer
+{
+    /// <summary>
+    /// Default Sql Connection Factory is responsible to handle Sql connections that can be made purely based on connection string.
+    /// Connection string containing user name and password, or integrated auth are perfect examples for this.
+    /// </summary>
+    public class DefaultSqlConnectionFactory : ISqlConnectionFactory
```
smithago, line 16: Where are the other implementations? We should keep all the implementations here so that when Fhir needs it when they move to AKS they will have it here easily

src/Microsoft.Health.SqlServer/Microsoft.Health.SqlServer.csproj
v-shaaal: Thanks for catching this. I added this earlier while trying something and forgot to remove it.

src/Microsoft.Health.SqlServer/DefaultSqlConnectionFactory.cs
v-shaaal: I kept ManagedIdentitySqlConnectionFactory in DICOM thinking that there may be other ways to fetch access token for different services. I can move it here and any service that needs a different implementation can override it.

src/Microsoft.Health.SqlServer/ISqlConnectionFactory.cs
v-shaaal: Updated as per our discussion offline.

src/Microsoft.Health.SqlServer/Configs/SqlServerConnectionType.cs
v-shaaal: Updated.

src/Microsoft.Health.SqlServer/Microsoft.Health.SqlServer.csproj
v-shaaal: Done.

src/Microsoft.Health.SqlServer/DefaultSqlConnectionFactory.cs
v-shaaal: Moved ManagedIdentitySqlConnection to shared components solutions. DefaultSqlConnection can be used for SqlUserNameAndPassword connection type and WindowsIntegratedAuth connection type as they both just rely on the connection string.

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            EnsureArg.IsNotNullOrEmpty(_sqlServerDataStoreConfiguration.ConnectionString);
+
+            SqlConnectionStringBuilder connectionBuilder = connectToMaster ?
+                new SqlConnectionStringBuilder(_sqlServerDataStoreConfiguration.ConnectionString) { InitialCatalog = string.Empty } :
+                new SqlConnectionStringBuilder(_sqlServerDataStoreConfiguration.ConnectionString);
+
+            SqlConnection sqlConnection = new SqlConnection(connectionBuilder.ToString());
+
+            AzureServiceTokenProvider azureServiceTokenProvider = new Microsoft.Azure.Services.AppAuthentication.AzureServiceTokenProvider();
```
jackliums, line 40: We should not be creating new instance of AzureServiceTokenProvider every time. This will result call to AAD to get new token every time a connection is created.

v-shaaal: Apologies, I missed to make these changes. I will make _azureServiceTokenProvider member of the class and access that. Since ManagedIdentitySqlConnection is registered as singleton service, _azureServiceTokenProvider should not be created every time. Please correct me if this understanding is not right.

tools/SchemaManager/Utils/BaseSchemaRunner.cs
```diff
@@ -4,7 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
-using Microsoft.Data.SqlClient;
+using System.Data.SqlClient;
```
smithago: I am not sure about this change, can you check with Brendon

v-shaaal: Microsoft.Data.SqlClient is now used throughout the solution. I believe it is Microsoft.Data.SqlClient in the latest iteration.

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+        private readonly AzureServiceTokenProvider _azureServiceTokenProvider;
+
+        public ManagedIdentityAccessTokenHandler()
+        {
+            _azureServiceTokenProvider = new AzureServiceTokenProvider();
+        }
+
+        public Task<string> GetAccessTokenAsync()
+        {
+            return _azureServiceTokenProvider.GetAccessTokenAsync("https://database.windows.net/");
```
brandonpollett, line 22: The value for the resource should probably be passed in for what you want to get a token for.

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.Health.SqlServer
+{
+    public class ManagedIdentityAccessTokenHandler : IAccessTokenHandler
+    {
+        private readonly AzureServiceTokenProvider _azureServiceTokenProvider;
+
+        public ManagedIdentityAccessTokenHandler()
+        {
+            _azureServiceTokenProvider = new AzureServiceTokenProvider();
```
brandonpollett, line 17: We will likely need to be able to pass in the other parameters. One case where this won't work is on things besides the public cloud.
See: https://github.com/Azure/azure-sdk-for-net/blob/6df900655ca8c4afe765ec847e6fab1fdc32193e/sdk/mgmtcommon/AppAuthentication/Azure.Services.AppAuthentication/AzureServiceTokenProvider.cs#L72

Some examples of possible values this could be:
https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud#azure-ad-authentication-endpoints

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
```diff
@@ -0,0 +1,54 @@
...
+            EnsureArg.IsNotNull(accessTokenHandler, nameof(accessTokenHandler));
+
+            _sqlServerDataStoreConfiguration = sqlServerDataStoreConfiguration;
+            _accessTokenHandler = accessTokenHandler;
+        }
+
+        /// <summary>
+        /// Get sql connection after getting access token.
+        /// </summary>
+        /// <param name="connectToMaster">Should connect to master?</param>
```
brandonpollett, line 30: I wonder if we should change the name of this to something like `clearInitialCatalog` to be more transparent as to what happens. Alternatively, we could change this to property to be `initialCatalog` and it could allow you to set it to whatever you wanted.

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -21,21 +21,24 @@ public class SqlServerHealthCheck : IHealthCheck
     {
         private readonly SqlServerDataStoreConfiguration _configuration;
         private readonly ILogger<SqlServerHealthCheck> _logger;
+        private readonly ISqlConnection _sqlConnection;
 
-        public SqlServerHealthCheck(SqlServerDataStoreConfiguration configuration, ILogger<SqlServerHealthCheck> logger)
+        public SqlServerHealthCheck(SqlServerDataStoreConfiguration configuration, ILogger<SqlServerHealthCheck> logger, ISqlConnection sqlConnection)
```
brandonpollett, line 26: nit: logger should be the last item.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -197,31 +198,38 @@ private bool CanInitialize()
...
+                            _logger.LogInformation("Created database");
+                        }
+                        else
+                        {
+                            _logger.LogWarning("Insufficient permissions to create the database");
+                            return false;
+                        }
 
-                    connection.Close();
+                        connection.Close();
```
brandonpollett, line 221: Since this is now within a using block, I don't believe the `connection.Close()` needs to be directly called anymore.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -85,6 +87,19 @@ public static class SqlServerBaseRegistrationExtensions
                 .AsSelf()
                 .AsImplementedInterfaces();
 
+            switch (config.ConnectionType)
+            {
+                case SqlServerConnectionType.ManagedIdentity:
+                    services.AddSingleton<ISqlConnection, ManagedIdentitySqlConnection>();
+                    services.AddSingleton<IAccessTokenHandler, ManagedIdentityAccessTokenHandler>();
+                    break;
+                case SqlServerConnectionType.WindowsIntegratedAuth:
```
brandonpollett, line 96: Is there any benefit to having this type since it's identical to `SqlUserNameAndPassword`? Is there a better name we could use to where there's only one of them?

tools/SchemaManager/SchemaManager.csproj
```diff
@@ -11,13 +11,18 @@
...
     <PackageReference Include="Ensure.That" Version="9.2.0" />
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.5.0" />
     <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="161.41011.9" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="Polly" Version="7.2.1" />
     <PackageReference Include="System.CommandLine.Experimental" Version="0.2.0-alpha.19174.3" />
     <PackageReference Include="System.CommandLine.Rendering" Version="0.2.0-alpha.19174.3" />
     <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.1" />
-    <PackageReference Include="Microsoft.Health.SqlServer" Version="1.0.0-master-20200826-1" />
+    <PackageReference Include="Microsoft.Health.SqlServer" Version="1.0.0-master-20200827-1" />
```
brandonpollett, line 21: We probably don't need this reference if it's being changed to a project reference.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -131,24 +134,23 @@ public static void ValidateDatabaseName(string databaseName)
...
-        public static SqlConnection GetConnectionIfDatabaseNotExists(string connectionString, string databaseName)
+        public static bool DoesDatabaseExist(SqlConnection masterDbConnection, string databaseName)
         {
-            // Connect to master database to evaluate if the requested database exists
-            var masterConnectionBuilder = new SqlConnectionStringBuilder(connectionString) { InitialCatalog = string.Empty };
-            var connection = new SqlConnection(masterConnectionBuilder.ToString());
-            connection.Open();
+            EnsureArg.IsNotNull(masterDbConnection, nameof(masterDbConnection));
+
+            masterDbConnection.Open();
```
brandonpollett, line 141: You probably need to check if the connection is already open here since I believe it will throw an exception if you try to open it again.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection.open?view=sqlclient-dotnet-core-2.0#Microsoft_Data_SqlClient_SqlConnection_Open

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -170,16 +172,15 @@ public static bool CreateDatabase(SqlConnection connection, string databaseName)
...
-            using (var connection = new SqlConnection(connectionString))
-            {
-                connection.Open();
+            EnsureArg.IsNotNull(connection, nameof(connection));
 
-                using (var command = new SqlCommand("SELECT count(*) FROM fn_my_permissions (NULL, 'DATABASE') WHERE permission_name = 'CREATE TABLE'", connection))
-                {
-                    return (int)command.ExecuteScalar() > 0;
-                }
+            connection.Open();
```
brandonpollett, line 179: Same here.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
v-shaaal: I agree. Both of them completely rely on connection string. I am lost for names though, completely open for suggestions.

tools/SchemaManager/SchemaManager.csproj
v-shaaal: Yep, will update.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
v-shaaal: Makes sense. Will update.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
v-shaaal: Noted.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
v-shaaal: I just double checked. What you are stating is absolutely correct. Will update in the next iteration.

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
v-shaaal: I agree. I will change it to `clearInitialCatalog` such that it can still fetch the initial catalog from connection string as before.

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
v-shaaal: yep, will update.

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
v-shaaal: Updated.

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
v-shaaal: Updated to accept `initialCatalog`.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -131,24 +136,26 @@ public static void ValidateDatabaseName(string databaseName)
             }
         }
 
-        public static SqlConnection GetConnectionIfDatabaseNotExists(string connectionString, string databaseName)
+        public static bool DoesDatabaseExist(SqlConnection masterDbConnection, string databaseName)
```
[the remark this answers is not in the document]
v-shaaal: Makes sense. Will update.

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Azure.Services.AppAuthentication;
+
+namespace Microsoft.Health.SqlServer
+{
+    public class ManagedIdentityAccessTokenHandler : IAccessTokenHandler
```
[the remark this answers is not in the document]
v-shaaal: I agree.

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
```diff
@@ -0,0 +1,67 @@
...
+            var result = await _accessTokenHandler.GetAccessTokenAsync(_azureResource);
+            sqlConnection.AccessToken = result;
+
+            return sqlConnection;
+        }
+
+        /// <inheritdoc />
+        public SqlConnection GetSqlConnection()
+        {
+            return GetSqlConnectionAsync().Result;
```
jackliums, line 58: Let's not do blocking async call.

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapperFactory.cs
```diff
@@ -4,34 +4,33 @@
...
         private readonly SqlTransactionHandler _sqlTransactionHandler;
         private readonly SqlCommandWrapperFactory _sqlCommandWrapperFactory;
+        private readonly ISqlConnectionFactory _sqlConnection;
 
         public SqlConnectionWrapperFactory(
-            SqlServerDataStoreConfiguration configuration,
             SqlTransactionHandler sqlTransactionHandler,
-            SqlCommandWrapperFactory sqlCommandWrapperFactory)
+            SqlCommandWrapperFactory sqlCommandWrapperFactory,
+            ISqlConnectionFactory sqlConnection)
```
jackliums, line 20: should this be sqlConnectionFactory?

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -17,55 +16,74 @@ public class SqlConnectionWrapper : IDisposable
...
             EnsureArg.IsNotNull(sqlTransactionHandler, nameof(sqlTransactionHandler));
             EnsureArg.IsNotNull(sqlCommandWrapperFactory, nameof(sqlCommandWrapperFactory));
+            EnsureArg.IsNotNull(sqlConnectionFactory, nameof(sqlConnectionFactory));
 
             _sqlTransactionHandler = sqlTransactionHandler;
             _enlistInTransactionIfPresent = enlistInTransactionIfPresent;
             _sqlCommandWrapperFactory = sqlCommandWrapperFactory;
+            _sqlConnectionFactory = sqlConnectionFactory;
+
+            InitializeAsync();
```
jackliums, line 39: You should not call async method from consturctor.

jackliums: The behavior is underministic because you are not waiting for the operation to finish before returning it so depending on how thread pool manages the tasks, it might not be completed before the caller access SqlConnectionproperty.

What I was suggesting is to make SqlConnectionWrapperFactory.ObtainSqlConnectionWrapper async call and within there, you can create an instance and also call InitializeAsync. I would make the constructor of SqlConnectionWrapper and InitializeAsync internal instead of public so that the life cycle management is completely done through SqlConnectionWrapperFactory which will always initializes it properly.

v-shaaal: Ahaan I misunderstood it earlier. This makes sense. I will make the changes and publish the updated iteration shortly.

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapperFactory.cs
```diff
@@ -4,34 +4,33 @@
...
             EnsureArg.IsNotNull(sqlCommandWrapperFactory, nameof(sqlCommandWrapperFactory));
+            EnsureArg.IsNotNull(sqlConnectionFactory, nameof(sqlConnectionFactory));
 
-            _configuration = configuration;
             _sqlTransactionHandler = sqlTransactionHandler;
             _sqlCommandWrapperFactory = sqlCommandWrapperFactory;
+            _sqlConnectionFactory = sqlConnectionFactory;
         }
 
         public SqlConnectionWrapper ObtainSqlConnectionWrapper(bool enlistInTransaction = false)
```
jackliums, line 31: This needs to be change to async

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -17,55 +16,74 @@ public class SqlConnectionWrapper : IDisposable
...
+            get { return _sqlConnection; }
+        }
+
+        public SqlTransaction SqlTransaction
+        {
+            get
+            {
+                return _sqlTransaction;
+            }
+        }
```
jackliums, line 51: should not do async void

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -18,33 +20,37 @@ namespace Microsoft.Health.SqlServer.Features.Schema
...
 
             _sqlServerDataStoreConfiguration = sqlServerDataStoreConfiguration;
             _schemaUpgradeRunner = schemaUpgradeRunner;
             _schemaInformation = schemaInformation;
+            _sqlConnectionFactory = sqlConnectionFactory;
             _logger = logger;
         }
 
-        public void Initialize(bool forceIncrementalSchemaUpgrade = false)
+        public async void InitializeAsync(bool forceIncrementalSchemaUpgrade = false)
```
jackliums, line 46: here too: avoid async void

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
v-shaaal: It's not really returning anything. What should the return type be?

v-shaaal: never mind, just the Task, right?

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
v-shaaal: Not sure if this can be changed as this is called in Start method which is basically an implementation of IStartable. Making Invoke method returning Task would require making Start method async which would require changes in IStartable, ultimately requiring changes in all the places where it is implemented.

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -17,55 +16,74 @@ public class SqlConnectionWrapper : IDisposable
...
 
-            if (_enlistInTransactionIfPresent && sqlTransactionHandler.SqlTransactionScope != null && sqlTransactionHandler.SqlTransactionScope.SqlConnection == null)
+            if (_enlistInTransactionIfPresent && _sqlTransactionHandler.SqlTransactionScope != null && _sqlTransactionHandler.SqlTransactionScope.SqlConnection == null)
             {
-                sqlTransactionHandler.SqlTransactionScope.SqlConnection = SqlConnection;
+                _sqlTransactionHandler.SqlTransactionScope.SqlConnection = SqlConnection;
             }
 
             if (SqlConnection.State != ConnectionState.Open)
             {
```
jackliums, line 71: this can be OpenAsync now

v-shaaal: True. I updated it before and with refactoring, I mistakenly undid my changes. Will fix.

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
```diff
@@ -0,0 +1,67 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Data.SqlClient;
+using Microsoft.Health.SqlServer.Configs;
+
+namespace Microsoft.Health.SqlServer
+{
+    public class ManagedIdentitySqlConnection : ISqlConnectionFactory
```
jackliums, line 13: Should this and DefaultSqlConnection ends with Factory?

src/Microsoft.Health.SqlServer/ISqlConnectionFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        /// <summary>
+        /// Get unopened SqlConnection object.
+        /// Initial catalog is determined from the connection string.
+        /// </summary>
+        /// <returns>SqlConnection object.</returns>
+        public Task<SqlConnection> GetSqlConnectionAsync();
```
jackliums, line 18: interface don't need modifiers (remove public)

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            EnsureArg.IsNotNull(azureServiceTokenProvider, nameof(azureServiceTokenProvider));
+
+            _azureServiceTokenProvider = azureServiceTokenProvider;
+        }
+
+        /// <inheritdoc />
+        public Task<string> GetAccessTokenAsync(string resource)
+        {
+            return _azureServiceTokenProvider.GetAccessTokenAsync(resource);
```
jackliums, line 26: GetAccessToken takes CancellationToken so we should honor that.

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnection.cs
v-shaaal: Yeah, I will update.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -85,6 +88,19 @@ public static class SqlServerBaseRegistrationExtensions
                 .AsSelf()
                 .AsImplementedInterfaces();
 
+            switch (config.AuthenticationType)
+            {
+                case SqlServerAuthenticationType.ManagedIdentity:
+                    services.AddSingleton<ISqlConnectionFactory, ManagedIdentitySqlConnectionFactory>();
+                    services.AddSingleton<IAccessTokenHandler, ManagedIdentityAccessTokenHandler>();
```
jackliums, line 95: If this is supposed to be reusabled, maybe this should always be registered

v-shaaal: Not necessarily. For now, we just fetch token if authentication type is managed identity. However, in future, if this is to be used for something else as well, we can register it without this check.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -194,31 +210,36 @@ private bool CanInitialize()
 
             if (_sqlServerDataStoreConfiguration.AllowDatabaseCreation)
             {
-                var connection = GetConnectionIfDatabaseNotExists(_sqlServerDataStoreConfiguration.ConnectionString, databaseName);
-
-                if (connection != null)
+                using (var connection = await _sqlConnectionFactory.GetSqlConnectionAsync(MasterDatabase))
```
jackliums, line 213: Is the new code actually equivalent of the previous code?

The previous code makes connection with InitialCatalog = string.Empty, this one is making connection with InitialCatalog = master?

v-shaaal: Yeah, this change was made after having discussion with Brandon.

v-shaaal: Main intention for this code is to connect to master and then check if database exists and then create database based on the result.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -232,7 +253,7 @@ public void Start()
         {
             if (!string.IsNullOrWhiteSpace(_sqlServerDataStoreConfiguration.ConnectionString))
             {
-                Initialize();
+                InitializeAsync().Wait();
```
smithago, line 256: can the start be async?

v-shaaal: We can have a separate user story to make this async. This Start is an implementation of the method in IStartable which is referenced in multiple files not related to sql server connection. Making this async would require changes in all those files as well. This would be absolutely a good idea to make them async at some point, sooner the better.

src/Microsoft.Health.SqlServer/ISqlConnectionFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        /// <returns>SqlConnection object.</returns>
+        Task<SqlConnection> GetSqlConnectionAsync();
+
+        /// <summary>
+        /// Get unopened SqlConnection object.
+        /// </summary>
+        /// <param name="initialCatalog">Initial catalog to connect to.</param>
+        /// <returns>SqlConnection object.</returns>
+        Task<SqlConnection> GetSqlConnectionAsync(string initialCatalog);
```
jackliums, line 25: Can these two just be collapsed into one? `Task<SqlConnection> GetSqlConnectionAsync(string initialCatalog = null);`?

v-shaaal: Absolutely. Will update shortly.

v-shaaal: Fixed.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
jackliums: The IStartable is right now now being called from IoC container which is sync call. There is a work item on this to make IStartable to be async but for now, I think this is fine.

smithago (approved):

src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
```diff
@@ -19,88 +19,88 @@ public class SchemaUpgradeRunner
...
             _scriptProvider = scriptProvider;
             _baseScriptProvider = baseScriptProvider;
-            _sqlServerDataStoreConfiguration = sqlServerDataStoreConfiguration;
             _mediator = mediator;
             _logger = logger;
+            _sqlConnectionFactory = sqlConnectionFactory;
         }
 
-        public void ApplySchema(int version, bool applyFullSchemaSnapshot)
+        public async Task ApplySchema(int version, bool applyFullSchemaSnapshot)
```
jackliums, line 46: Keep the naming consistent. ApplySchemaAsync and ApplyBaseSchemaAsync

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -17,55 +17,72 @@ public class SqlConnectionWrapper : IDisposable
...
-                SqlConnection.Open();
+                await SqlConnection.OpenAsync();
```
jackliums, line 72: Most of the Async calls takes CancellationToken so we need to honor them.

jackliums (commented):
A few minor comments with CancellationToken and naming convention.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
v-shaaal: A fix for this is made in [AB#106](https://github.com/microsoft/healthcare-shared-components/pull/106).

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
v-shaaal: A fix for this is made in [AB#106](https://github.com/microsoft/healthcare-shared-components/pull/106).

PR #103, opened 2020-09-28
Remove connection string from logging

abdullah248:
## Description
Remove connection string from logging during initialization of Blob storage.

## Related issues
https://microsofthealth.visualstudio.com/Health/_workitems/edit/76244

## Testing
Ran locally with local version of dicom-server to ensure connection string not logged.


brendankowitz (approved):
😬

PR #108, opened 2020-10-12
Adding in Option for MI with Blob Storage

snehabandla:
## Description
Adding in an option for MI in Blob Storage Config

## Related issues
[User Story 76203](https://microsofthealth.visualstudio.com/Health/_workitems/edit/76203)

## Testing
Manually tested

### DIFF
--- a/src/Microsoft.Health.Blob/Features/Storage/BlobClientInitializer.cs
+++ b/src/Microsoft.Health.Blob/Features/Storage/BlobClientInitializer.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Azure.Identity;
 using Azure.Storage.Blobs;
 using EnsureThat;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Blob/Features/Storage/BlobClientInitializer.cs
```diff
@@ -41,6 +42,12 @@ public BlobServiceClient CreateBlobClient(BlobDataStoreConfiguration configurati
             blobClientOptions.Retry.Delay = TimeSpan.FromSeconds(configuration.RequestOptions.ExponentialRetryBackoffDeltaInSeconds);
             blobClientOptions.Retry.NetworkTimeout = TimeSpan.FromMinutes(configuration.RequestOptions.ServerTimeoutInMinutes);
 
+            if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
+            {
+                var defaultCredentials = new DefaultAzureCredential();
+                return new BlobServiceClient(new Uri(configuration.ConnectionString), defaultCredentials, blobClientOptions);
```
smithago, line 48: FYI, Brandon raised this before, for managed identity, the ConnectionString is actually just the blob service url. For now it is okay to overload the same setting. We should document it in Dicom atleast to show how it looks for managed identity

smithago (approved):

PR #110, opened 2020-10-12
Fixes base schema execution

rbans96:
## Description
Updates base schema for async and fixes the base schema execution by tool while any sql connection is open by the server.

## Related issues
Addresses [issue #].

## Testing
Ran the schema migration tool with schema as null to ensure that base schema is executed before upgrading the schema.

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
@@ -5,8 +5,9 @@
     Configure database
 **************************************************************/
 -- Enable RCSI
+-- ROLLBACK IMMEDIATE will cut off live connections and roll back any open transactions immediately
 IF ((SELECT is_read_committed_snapshot_on FROM sys.databases WHERE database_id = DB_ID()) = 0) BEGIN
-    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON
+    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON WITH ROLLBACK IMMEDIATE
 END
 
 -- Avoid blocking queries when statistics need to be rebuilt
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
@@ -5,9 +5,8 @@
     Configure database
 **************************************************************/
 -- Enable RCSI
--- ROLLBACK IMMEDIATE will cut off live connections and roll back any open transactions immediately
 IF ((SELECT is_read_committed_snapshot_on FROM sys.databases WHERE database_id = DB_ID()) = 0) BEGIN
-    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON WITH ROLLBACK IMMEDIATE
+    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON
 END
 
 -- Avoid blocking queries when statistics need to be rebuilt
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
@@ -4,10 +4,10 @@
 /*************************************************************
     Configure database
 **************************************************************/
-
 -- Enable RCSI
+-- ROLLBACK IMMEDIATE will cut off live connections and roll back any open transactions immediately
 IF ((SELECT is_read_committed_snapshot_on FROM sys.databases WHERE database_id = DB_ID()) = 0) BEGIN
-    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON
+    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON WITH ROLLBACK IMMEDIATE
 END
 
 -- Avoid blocking queries when statistics need to be rebuilt
--- a/tools/SchemaManager/Commands/ApplyCommand.cs
+++ b/tools/SchemaManager/Commands/ApplyCommand.cs
@@ -57,7 +57,7 @@ namespace SchemaManager.Commands
                     {
                         Console.WriteLine(string.Format(Resources.RetryCurrentSchemaVersion, attemptCount++, RetryAttempts));
                     })
-                .ExecuteAsync(() => FetchUpdatedAvailableVersionsAsync(schemaClient, connectionString, cancellationToken));
+                .ExecuteAsync(token => FetchUpdatedAvailableVersionsAsync(schemaClient, connectionString, token), cancellationToken);
 
                 if (availableVersions.Count == 1)
                 {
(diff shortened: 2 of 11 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
```diff
@@ -4,10 +4,10 @@
 /*************************************************************
     Configure database
 **************************************************************/
-
 -- Enable RCSI
+-- ROLLBACK IMMEDIATE will cut off live connections and roll back any open transactions immediately
 IF ((SELECT is_read_committed_snapshot_on FROM sys.databases WHERE database_id = DB_ID()) = 0) BEGIN
-    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON
+    ALTER DATABASE CURRENT SET READ_COMMITTED_SNAPSHOT ON WITH ROLLBACK IMMEDIATE
```
rbans96, line 10: Since this will be executed only once when the schema is null, so I think there will not be any transactions running at this point and its safe to ensure that any connection by server is closed.

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -150,12 +150,12 @@ await Policy.Handle<SchemaManagerException>()
             }
         }
 
-        private static void UpgradeSchema(string connectionString, int version, string script)
+        private static async Task UpgradeSchema(string connectionString, int version, string script)
```
v-shaaal, line 153: Should rename it to UpgradeSchemaAsync for consistency.

src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
rotodd: Tested this locally with the test web project and I'm no longer seeing any delays in the base schema execution, and the `apply --next` command succeeds 👍.

I did a bit of reading about `WITH ROLLBACK IMMEDIATE` and found: "this option will rollback all uncommitted transactions and reset all connections" (from [stackoverflow](https://stackoverflow.com/questions/34589660/sql-server-alter-database-with-rollback-immediate)). Did the setup for SQL MI create a connection that was blocking the base schema from being excuted? Just trying to figure out why there is now an open connection that we need to reset.

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -150,39 +150,39 @@ await Policy.Handle<SchemaManagerException>()
...
+            CompatibleVersion compatibleVersion = await schemaClient.GetCompatibilityAsync(cancellationToken);
 
             if (maxAvailableVersion > compatibleVersion.Max)
             {
                 throw new SchemaManagerException(string.Format(Resources.VersionIncompatibilityMessage, maxAvailableVersion));
             }
         }
 
-        private static async Task ValidateInstancesVersion(ISchemaClient schemaClient, int version)
+        private static async Task ValidateInstancesVersion(ISchemaClient schemaClient, int version, CancellationToken cancellationToken)
```
rotodd, line 183: Should this also have "async" in the name to be consistent?

src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
YazanMSFT: If someone were running an older version of the code with schema version 1 or 2, and upgraded the FHIR service directly to the latest and ran schema migration, would this script get run?

rbans96: @YazanMSFT  baseSchema.sql would be executed but not alter database scripts since if condition would not evaluate to true.

YazanMSFT: Thanks, that makes sense. We do have the option of specifying WITH ROLLBACK AFTER [n] rather than immediate, to give a chance for stuff to complete before forcing them to rollback. But if we're sure it won't get run then immediate is fine.

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -114,12 +114,12 @@ await Policy.Handle<SchemaManagerException>()
                             {
                                 Console.WriteLine(string.Format(Resources.RetryCurrentVersions, attemptCount++, RetryAttempts));
                             })
-                        .ExecuteAsync(() => ValidateInstancesVersion(schemaClient, executingVersion));
+                        .ExecuteAsync(() => ValidateInstancesVersion(schemaClient, executingVersion, cancellationToken));
```
YazanMSFT, line 117: Thanks for adding cancellation tokens! ExecuteAsync has an overload that accepts one too, which should prevent it from continuing to retry when cancellation is requested.

tools/SchemaManager/Utils/BaseSchemaRunner.cs
```diff
@@ -40,25 +40,25 @@ public static async Task EnsureBaseSchemaExistsAsync(string connectionString, Ca
...
+            await Policy.Handle<SchemaManagerException>()
+            .WaitAndRetryAsync(
                 retryCount: RetryAttempts,
                 sleepDurationProvider: (retryCount) => RetrySleepDuration,
                 onRetry: (exception, retryCount) =>
                 {
                     Console.WriteLine(string.Format(Resources.RetryInstanceSchemaRecord, attempts++, RetryAttempts));
                 })
-            .Execute(() => InstanceSchemaRecordCreated(connectionString));
+            .ExecuteAsync(() => InstanceSchemaRecordCreatedAsync(connectionString, cancellationToken));
```
YazanMSFT, line 56: Similarly here, ExecuteAsync can take a token too.

tools/SchemaManager/SchemaDataStore.cs
```diff
@@ -84,58 +86,58 @@ public static int GetCurrentSchemaVersion(string connectionString)
...
             }
         }
 
-        public static void ExecuteScript(string connectionString, string script)
+        public static async Task ExecuteScript(string connectionString, string script)
         {
             using (var connection = new SqlConnection(connectionString))
             {
-                connection.Open();
+                await connection.OpenAsync();
```
YazanMSFT, line 105: Nit: this can take a cancellation token.

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -57,7 +57,7 @@ await Policy.Handle<SchemaManagerException>()
                     {
                         Console.WriteLine(string.Format(Resources.RetryCurrentSchemaVersion, attemptCount++, RetryAttempts));
                     })
-                .ExecuteAsync(() => FetchUpdatedAvailableVersions(schemaClient, connectionString));
+                .ExecuteAsync(() => FetchUpdatedAvailableVersionsAsync(schemaClient, connectionString, cancellationToken));
```
YazanMSFT, line 60: Similarly here.

src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
rbans96: Good point. I initially gave it a thought before creating this PR to ROLLBACK AFTER 30 seconds or so. But then realizing that since schema would be null at this point so no transactions should ideally run.

tools/SchemaManager/Commands/ApplyCommand.cs
rbans96: Thanks, fixed.

YazanMSFT (approved):

src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql
rbans96: @rotodd Good question. I took a look at this [PR](https://github.com/microsoft/healthcare-shared-components/pull/102) and seems connection is scoped to using block so not sure what's causing the open connection. But I think IMMEDIATE ROLLBACK is okay to put in since no schema would exist at this point(and more comments above)

PR #111, opened 2020-10-13
Audit Egress Logger

v-shaaal:
## Description
This pull request introduces the concept of Audit Egress Logger. Both FHIR and DICOM handle auditing of executed messages differently. Audit Egress Logger provides both of them the opportunity to write log executed messages as per their respective requirements.

**Details:**
**FHIR** logs executed messages in two different places. First in AuditLoggingFilterAttribute, and second in AuditMiddleware. AuditMiddleware was specifically taking care of AuthX failure messages. All the logging messages in FHIR cannot be moved to AuditMiddleware because of batch statements. AuditLoggingFilterAttribute helps with writing log executed messages for each sub request present in the batch request. If this is moved to AuditMiddleware, we will end up writing just one executed message for a batch which is not the expected behavior.
**DICOM** will now log all the executed messages in AuditMiddleware using AuditEgressLogger. Since the concept of batch does not exist in DICOM, all the executed messages can be written in AuditMiddleware which will save us from writing executed messages in two different places.

## Related Issues
[AB#76579](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76579).

## Testing
All the existing tests are passing.


src/Microsoft.Health.Api/Features/Audit/IAuditEgressLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Health.Core.Features.Security;
+
+namespace Microsoft.Health.Api.Features.Audit
+{
+    /// <summary>
+    /// Logger to log executed messages.
+    /// </summary>
+    public interface IAuditEgressLogger
```
smithago, line 14: I am not sure how this helps Ali. There is still 2 interfaces methods with LogExcuted, this one and IAuditEgressLogger

v-shaaal: Deleted IAuditEgressLogger. LogExecuted logic will be moved to AuditHelper implementation in FHIR and DICOM as needed.

smithago (approved):

Issue #113, opened 2020-10-16
Eliminate need to insert into SchemaVersion table at start of "full" migration script


johnstairs:
We should be able to eliminate this from the start of every 6.sql, 7.sql, etc file:

```SQL
/*************************************************************
    Schema Version
**************************************************************/

INSERT INTO dbo.SchemaVersion
VALUES
    (6, 'started')

GO
```

Instead, we could insert the row before calling the script, like we do for .diff.sql scripts.

rbans96:
Created #[AB76682](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=76682).

PR #117, opened 2020-10-19
Eliminate the need to insert schema version from 6.sql, 7.sql..

rbans96:
## Description
Inserts the schema version as started for full scripts(6.sql, 7.sql..) in the same manner as we insert for diff scripts.

## Related issues
Addresses #113 

## Testing
Tested manually by switching the AutomaticUpdatesEnabled flag to true/false

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInformation.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInformation.cs
@@ -18,5 +18,14 @@ namespace Microsoft.Health.SqlServer.Features.Schema
         public int MaximumSupportedVersion { get; }
 
         public int? Current { get; set; }
+
+        /// <summary>
+        /// Checks if the version is not inside the full schema script.
+        /// </summary>
+        /// <param name="version">The schema version.</param>
+        public static bool DoesNotContainsVersionInSchema(int version)
+        {
+            return version > 5;
+        }
     }
 }
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Health.SqlServer.Features.Schema
         {
             _logger.LogInformation("Applying schema {version}", version);
 
-            // Eliminate the need to insert schema version from 6.sql, 7.sql and so on.
+            // Insert schema version as started for diff scripts and 6.sql, 7.sql and so on.
             if (!applyFullSchemaSnapshot || version > 5)
             {
                 await InsertSchemaVersionAsync(version, cancellationToken);
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
@@ -48,7 +48,8 @@ namespace Microsoft.Health.SqlServer.Features.Schema
         {
             _logger.LogInformation("Applying schema {version}", version);
 
-            if (!applyFullSchemaSnapshot)
+            // Eliminate the need to insert schema version from 6.sql, 7.sql and so on.
+            if (!applyFullSchemaSnapshot || version > 5)
             {
                 await InsertSchemaVersionAsync(version, cancellationToken);
             }
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Health.SqlServer.Features.Schema
             _logger.LogInformation("Applying schema {version}", version);
 
             // Insert schema version as started for diff scripts and 6.sql, 7.sql and so on.
-            if (!applyFullSchemaSnapshot || version > 5)
+            if (!applyFullSchemaSnapshot || SchemaInformation.DoesNotContainsVersionInSchema(version))
             {
                 await InsertSchemaVersionAsync(version, cancellationToken);
             }
--- a/tools/SchemaManager/Commands/ApplyCommand.cs
+++ b/tools/SchemaManager/Commands/ApplyCommand.cs
@@ -152,6 +152,7 @@ namespace SchemaManager.Commands
 
         private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken, bool isFullSchemaSnapshot = false)
         {
+            // Replace the check(version > 5) with SchemaInformation.DoesNotContainsVersionInSchema after Microsoft.Health.SqlServer dependency upgrade.
             if (!isFullSchemaSnapshot || version > 5)
             {
                 await SchemaDataStore.UpsertSchemaVersionAsync(connectionString, version, "started", cancellationToken);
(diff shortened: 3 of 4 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
```diff
@@ -48,7 +48,8 @@ public async Task ApplySchemaAsync(int version, bool applyFullSchemaSnapshot, Ca
         {
             _logger.LogInformation("Applying schema {version}", version);
 
-            if (!applyFullSchemaSnapshot)
+            // Eliminate the need to insert schema version from 6.sql, 7.sql and so on.
```
Ivanidzo4ka, line 51: > // Eliminate the need to insert schema version from 6.sql, 7.sql and so on. [](start = 11, length = 76)

Any chance you can rephrase comment here from instead of what it is doing to why it is doing it?

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -150,10 +150,17 @@ await Policy.Handle<SchemaManagerException>()
             }
         }
 
-        private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken)
+        private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken, bool isFullSchemaSnapshot = false)
         {
-            // check if the record for given version exists in failed status
-            await SchemaDataStore.DeleteSchemaVersionAsync(connectionString, version, SchemaDataStore.Failed, cancellationToken);
+            if (!isFullSchemaSnapshot || version > 5)
```
Ivanidzo4ka, line 155: >5 [](start = 51, length = 1)

Is there are place in code where you can have `static const int SchemaWithVersion ` = 5 ?Two different places in code rely on same constant, it's easy to make mistake, or lost track of them

src/Microsoft.Health.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
rbans96: Thanks, fixed.

tools/SchemaManager/Commands/ApplyCommand.cs
rbans96: Good point! I don't think we have any common place for the tool and sqlServer to define static constants. This schema version '5' is added since we are changing the approach to insert schema version for full scripts for schema version>5. I don't think changing this later but we might plan to take this(version > 5) out when we update minSupportedVersion>5. So not sure,  if I should create a separate class just for this one to be shared. Thoughts?

Ivanidzo4ka: Can be totally wrong, but I would create static method in `SchemaInformation` class
something like:

```
///<summary>Checks is provided version already has schema version inside schema script.</summary>
public static bool ContainsVersionInSchema(int version) =>version>5;
```
and use it in both places.


---
In reply to: [509749340](https://github.com/microsoft/healthcare-shared-components/pull/117#discussion_r509749340) [](ancestors = 509749340)

Ivanidzo4ka (approved):
:shipit:

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -150,10 +150,18 @@ await Policy.Handle<SchemaManagerException>()
             }
         }
 
-        private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken)
+        private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken, bool isFullSchemaSnapshot = false)
         {
-            // check if the record for given version exists in failed status
-            await SchemaDataStore.DeleteSchemaVersionAsync(connectionString, version, SchemaDataStore.Failed, cancellationToken);
+            // Replace the check(version > 5) with SchemaInformation.DoesNotContainsVersionInSchema after Microsoft.Health.SqlServer dependency upgrade.
+            if (!isFullSchemaSnapshot || version > 5)
```
Ivanidzo4ka, line 156: >version > 5 [](start = 41, length = 11)

Don't you want to use `SchemaInformation.DoesNotContainsVersionInSchema(version)` here?

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInformation.cs
```diff
@@ -18,5 +18,14 @@ public SchemaInformation(int minimumSupportedVersion, int maximumSupportedVersio
         public int MaximumSupportedVersion { get; }
 
         public int? Current { get; set; }
+
+        /// <summary>
+        /// Checks if the version is not inside the full schema script.
+        /// </summary>
+        /// <param name="version">The schema version.</param>
+        public static bool DoesNotContainsVersionInSchema(int version)
```
rotodd, line 26: Nit: what about `SetsVersionInSchema` and `return version <= 5` (just to remove the negative from the boolean name)

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInformation.cs
```diff
@@ -18,5 +18,14 @@ public SchemaInformation(int minimumSupportedVersion, int maximumSupportedVersio
         public int MaximumSupportedVersion { get; }
 
         public int? Current { get; set; }
+
+        /// <summary>
+        /// Checks if the version is not inside the full schema script.
```
rotodd, line 23: It might be good to add a bit more context here about how previous full snapshot schemas set the schema version and that we are looking to prevent future SQL schemas from having to do so.

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -150,10 +150,18 @@ await Policy.Handle<SchemaManagerException>()
             }
         }
 
-        private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken)
+        private static async Task UpgradeSchemaAsync(string connectionString, int version, string script, CancellationToken cancellationToken, bool isFullSchemaSnapshot = false)
```
rotodd, line 153: Just curious, why did you set `isFullSchemaSnapshot` to default to `false`?

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -150,10 +150,18 @@ await Policy.Handle<SchemaManagerException>()
...
         {
-            // check if the record for given version exists in failed status
-            await SchemaDataStore.DeleteSchemaVersionAsync(connectionString, version, SchemaDataStore.Failed, cancellationToken);
+            // Replace the check(version > 5) with SchemaInformation.DoesNotContainsVersionInSchema after Microsoft.Health.SqlServer dependency upgrade.
+            if (!isFullSchemaSnapshot || version > 5)
+            {
+                await SchemaDataStore.UpsertSchemaVersionAsync(connectionString, version, "started", cancellationToken);
```
rotodd, line 158: Nit: it might make sense to make `"started"` a Resource or a `const`

rotodd:
+1 looks good! Really awesome to have this automated 😎

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInformation.cs
```diff
@@ -18,5 +18,14 @@ public SchemaInformation(int minimumSupportedVersion, int maximumSupportedVersio
...
+        /// </summary>
+        /// <param name="version">The schema version.</param>
+        public static bool DoesNotContainsVersionInSchema(int version)
+        {
+            return version > 5;
+        }
```
rotodd, line 29: Coming back to this PR after thinking about it for a while - should `version > 5` be hard coded in this repo if this bool is specific to how the schema files are set up in the fhir-server repo?

rbans96:
Working on having a better approach to have this change on FHIR/DICOM. I will create a separate PR with the updated approach.

PR #118, opened 2020-10-20
Implement IHostedService instead of IStartable

johnstairs:
## Description
Removing the synchronous `IStartable` interface. In its place, we implement the built-in `IHostedService`, which is asynchronous and is invoked by the `Host` after the DI container has been built.

## Testing
I've implemented the changes on the fhir server and have tests passing.


test/Microsoft.Health.SqlServer.Web/Properties/launchSettings.json
```diff
@@ -3,11 +3,7 @@
     "Microsoft.Health.SqlServer.Web": {
       "commandName": "Project",
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "development",
```
johnstairs, line 6: @rbans96 , I removed these environment variables to disable the `SchemaInitializer` from running in the E2E test server. Previously, the schema initializer did not run because we did not register the `InitializationModule`, so `IStartable`s were not started. Now with change, `IHostedService`s *will* be started. When running in Github actions, we do not have a database to connect to at the given connection string so it fails.

src/Microsoft.Health.Api.UnitTests/Features/Audit/AuditEventTypeMappingTests.cs
```diff
@@ -64,7 +65,7 @@ public AuditEventTypeMappingTests()
 
             _auditEventTypeMapping = new AuditEventTypeMapping(_actionDescriptorCollectionProvider);
 
-            ((IStartable)_auditEventTypeMapping).Start();
+            ((IHostedService)_auditEventTypeMapping).StartAsync(CancellationToken.None);
```
smithago, line 68: Ideally we would await for a async task to finish, unless it is fire and forget. Should we move this out of the constructor?

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -31,7 +33,7 @@ namespace Microsoft.Health.Api.Features.Audit
     /// This class builds the mapping ahead of time so that we can lookup the audit event type any time during the pipeline given the controller name and action name.
     /// </para>
     /// </remarks>
-    public class AuditEventTypeMapping : IAuditEventTypeMapping, IStartable
+    public class AuditEventTypeMapping : IAuditEventTypeMapping, IHostedService
```
smithago, line 36: My only thought with using IHostedService is, are we overloading the meaning of this Interface.  The name is throwing me off. 
Looks like eventhough we need same interface, IHostedService is furture used in BackGroundServices.

johnstairs: I had a similar thought. Here is a counterpoint: the term `service` is heavily overloaded. One meaning is in the sense of a DI service (`IServiceProvider`). This is a special kind of such a service, one that the runtime explicitly instantiates, starts, and stops. It does not matter whether it continues to do anything in the background in between.

If that is not convincing, we can create an `IHostedService` that instantiates, starts and stops instances of another kind of service (`IStartStoppable` :)) with basically the same signature. But the indirection does not seem very helpful. Also, with IHostedService it is clear when StartAsync and StopAsync will be called (since it is [documented](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostedservice?view=dotnet-plat-ext-3.1)) and we are not inventing our own concept.

brendankowitz:
This PR is pretty scary for me. IStartable has a guarantee that it will happen after the construction of the IoC container, but before app initialization is complete. This PR changes that behavior to async background tasks that complete at unknown intervals

src/Microsoft.Health.Api/Modules/InitializationModule.cs
```diff
@@ -38,17 +36,6 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
...
         private static void Configure(IApplicationBuilder app)
         {
-            ILogger logger = app.ApplicationServices.GetRequiredService<ILoggerFactory>().CreateLogger<InitializationModule>();
-
-            // start IStartable services.
-            foreach (var startable in app.ApplicationServices.GetService<IEnumerable<IStartable>>())
-            {
-                using (logger.BeginTimedScope($"Initializing {startable.GetType().Name}."))
-                {
-                    startable.Start();
```
brendankowitz, line 48: This PR is pretty scary for me. IStartable has a guarantee that it will happen after the construction of the IoC container, but before app initialization is complete. This PR changes that behavior to async background tasks that complete at unknown intervals. 

I think this behavior is also assumed by components that implement `IRequireInitializationOnFirstRequest`, where they expect the Start() to have completed, here that is no longer guaranteed.

johnstairs: `StartAsync`s are awaited before the Host is considered "started" 
https://github.com/dotnet/extensions/blob/1264b9e2567027d2243f77f48cc08fdd576e3005/src/Hosting/Hosting/src/Internal/Host.cs#L47-L51

I don't think it's all that different from what we had before.

johnstairs: I put a 30s delay in `StartAsync` and the there is no response to the web request. So I think we should be ok.

test/Microsoft.Health.SqlServer.Web/Properties/launchSettings.json
rbans96: Thanks for the information.

test/Microsoft.Health.SqlServer.Web/Properties/launchSettings.json
```diff
@@ -3,11 +3,7 @@
...
-        "SqlServer:Initialize": "true",
```
rbans96, line 7: I am wondering if we could make Initialize and allowDatabaseCreation as false and not remove these. And when we plan to create a database to connect to, we can flip them on. I use these settings frequently for my local testing so just gave it a thought but it could be just me :)

johnstairs: Sure!

src/Microsoft.Health.Api.UnitTests/Features/Audit/AuditEventTypeMappingTests.cs
johnstairs: Done.

test/Microsoft.Health.SqlServer.Web/Properties/launchSettings.json
rotodd: To make sure I understand, if we would like to perform an auto schema upgrade for an existing database, we will need to flip `SqlServer:Initialize` to be true, but if we would like to perform a manual schema upgrade, this won't be needed - correct?

If so, it might be worthwhile to make an investigate story and dig into these config options a bit more. It would be good to ensure their functionality is clear and consistent.

src/Microsoft.Health.Api/Modules/InitializationModule.cs
brendankowitz: Thanks for verifying John!

brendankowitz (approved):

src/Microsoft.Health.Api/Modules/InitializationModule.cs
johnstairs: It's actually simpler than that: they add the `GenericWebHostService` last so all other services are started first:
image

PR #120, opened 2020-10-29
Throw specific exception when fail to retrieve token

abdullah248:
## Description
When the clients failed to obtain a token it would just throw a System.ArgumentNullException. This updates to throw a specific FailToRetrieveTokenException in the case that it does not obtain a token in the OAuth2ClientCredentialProvider or the OAuth2UserPasswordCredentialProvider.

## Related issues
https://microsofthealth.visualstudio.com/Health/_boards/board/t/Medical%20Imaging/Stories/?workitem=74886

## Testing
Added unit tests.

### DIFF
--- a/src/Microsoft.Health.Client/OAuth2ClientCredentialProvider.cs
+++ b/src/Microsoft.Health.Client/OAuth2ClientCredentialProvider.cs
@@ -49,9 +49,9 @@ namespace Microsoft.Health.Client
             using HttpResponseMessage tokenResponse = await _httpClient.PostAsync(_oAuth2ClientCredentialConfiguration.TokenUri, formContent, cancellationToken);
 
             var openIdConnectMessage = new OpenIdConnectMessage(await tokenResponse.Content.ReadAsStringAsync());
-            if (openIdConnectMessage.AccessToken == null)
+            if (tokenResponse.StatusCode == System.Net.HttpStatusCode.BadRequest)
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Client/OAuth2ClientCredentialProvider.cs
```diff
@@ -48,6 +49,11 @@ protected override async Task<string> BearerTokenFunction(CancellationToken canc
             using HttpResponseMessage tokenResponse = await _httpClient.PostAsync(_oAuth2ClientCredentialConfiguration.TokenUri, formContent, cancellationToken);
 
             var openIdConnectMessage = new OpenIdConnectMessage(await tokenResponse.Content.ReadAsStringAsync());
+            if (openIdConnectMessage.AccessToken == null)
```
smithago, line 52: Should we check this or should we check the httpClient response status code?

smithago (approved):

src/Microsoft.Health.Client/OAuth2ClientCredentialProvider.cs
abdullah248: Updated to check the response code

smithago: can we move this above openIdConnectMessage

PR #125, opened 2020-11-14
Adds Semver to Assemblies, Nugets and Release names

brendankowitz:
## Description
Adds semver with [GitVersioning](https://github.com/GitTools/GitVersion)

This package allows controlling versioning through Git, version bumps can be utilized through `+semver: major|minor|feature|breaking` syntax in commit messages. 

For all incremental bug fixes, nothing needs to be changed, GitVersion will increment the patch version automatically based on commits in main.

For breaking changes you can use the following syntax in the commit message:
```
+semver: breaking
or
+semver: major
```

Features can use:
```
+semver: feature
or
+semver: minor
```

I believe this will help solve a lot of the incompatibilities with breaking changes we've been seeing with downstream projects.

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -2,14 +2,18 @@ parameters:
   packageArtifacts: true
   
 steps:
+- template: ./update-semver.yml
 
 - task: UseDotNet@2
   displayName: 'Use .NET Core sdk'
   inputs:
     useGlobalJson: true
 
-- script: dotnet build --configuration $(buildConfiguration) --version-suffix $(build.buildNumber) /warnaserror
+- task: DotNetCoreCLI@2
   displayName: 'dotnet build $(buildConfiguration)'
+  inputs:
+    command: build
+    arguments: '--configuration $(buildConfiguration) /p:AssemblyVersion="$(assemblySemVer)" /p:FileVersion="$(assemblySemFileVer)" /p:InformationalVersion="$(informationalVersion)" /p:Version="$(majorMinorPatch)" /warnaserror'
 
 - task: DotNetCoreCLI@2
   displayName: 'dotnet test UnitTests'
--- a/build/build.yml
+++ b/build/build.yml
@@ -13,7 +13,7 @@ steps:
   displayName: 'dotnet build $(buildConfiguration)'
   inputs:
     command: build
-    arguments: '--configuration $(buildConfiguration) /p:AssemblyVersion="$(assemblySemVer)" /p:FileVersion="$(assemblySemFileVer)" /p:InformationalVersion="$(informationalVersion)" /p:Version="$(majorMinorPatch)" /warnaserror'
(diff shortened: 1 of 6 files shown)


feordin:
Just to confirm, we will add the +semver: <type> text to the squash and merge commit that goes into master when we close our PR, correct?

brendankowitz:
> Just to confirm, we will add the +semver: text to the squash and merge commit that goes into master when we close our PR, correct?

Yes, it will look in the commit messages in the master branch so it needs to be in that final commit. 
Its fine to have these also in a branch and you can see how the versioning will be applied

brandonpollett (approved):

docs/Versioning.md
```diff
@@ -0,0 +1,35 @@
+# Semantic Versioning for Shared Components
+
+This guide gives an overview of the Semantic versioning implementation in use with this project.
+To achieve semantic versioning consistently and relaiably the [GitVersion](https://github.com/GitTools/GitVersion) library is used.
```
johnstairs, line 4: typo

build/build.yml
```diff
@@ -2,14 +2,18 @@ parameters:
...
   displayName: 'Use .NET Core sdk'
   inputs:
     useGlobalJson: true
 
-- script: dotnet build --configuration $(buildConfiguration) --version-suffix $(build.buildNumber) /warnaserror
+- task: DotNetCoreCLI@2
   displayName: 'dotnet build $(buildConfiguration)'
+  inputs:
+    command: build
+    arguments: '--configuration $(buildConfiguration) /p:AssemblyVersion="$(assemblySemVer)" /p:FileVersion="$(assemblySemFileVer)" /p:InformationalVersion="$(informationalVersion)" /p:Version="$(majorMinorPatch)" /warnaserror'
```
johnstairs, line 16: When we change the assembly version, the code will not be binary-compatible. A recompilation of the consuming app will be required. 
So we should maybe be clear that even in cases were we only increment the patch, we will not be binary-compatible.

brendankowitz: We could potentially trim this to have a fixed Minor.Patch, e.g. {Major}.0.0? Perhaps this makes more sense.
I think leaving the informational version as-is should be fine to inspect the origin

johnstairs: The problem is that it's really easy to introduce changes that require recompilation but that are binary compatible. So I don't think we should do this. 
By the way, what is out level of commitment to not making breaking changes without updating the major number? Should we have a tool that checks the API surface?

brendankowitz: Makes sense, I'll revert the version string to a hard-coded "1.0.0" but keep the informational version.

We are also not publishing any packages to the official nuget feed yet, so I see some opportunity to keep improving here. I like the suggestion, do you know of any tools already? I'll can look into this

brendankowitz: With changes from above:
image

johnstairs (approved):

PR #129, opened 2020-11-26
Removes "SqlServer:Initialize" config option from launch settings

rotodd:
## Description
Now that we have `SqlServer:AutomaticUpdatesEnabled`, the `SqlServer:Initialize` config setting is no longer needed.

This PR removes `SqlServer:Initialize` and updates an error message for an edge case that is a bit easier to trigger now that we are removing this setting (please see my review comment below for more context).

## Related issues
Addresses [#AB77211](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77211).

## Testing
Manually tested the startup of the test web project with all possible combinations of the config values. I also copied the SQL health .pdb and .dll files and tested the fhir-server startup as well.

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
@@ -205,16 +205,16 @@ namespace Microsoft.Health.SqlServer.Features.Schema
 
             ValidateDatabaseName(databaseName);
 
-            if (_sqlServerDataStoreConfiguration.AllowDatabaseCreation)
+            using (var connection = await _sqlConnectionFactory.GetSqlConnectionAsync(MasterDatabase, cancellationToken))
             {
-                using (var connection = await _sqlConnectionFactory.GetSqlConnectionAsync(MasterDatabase, cancellationToken))
+                bool doesDatabaseExist = await DoesDatabaseExistAsync(connection, databaseName, cancellationToken);
+
+                if (!doesDatabaseExist)
                 {
-                    bool doesDatabaseExist = await DoesDatabaseExistAsync(connection, databaseName, cancellationToken);
+                    _logger.LogInformation("Database does not exist");
 
-                    if (!doesDatabaseExist)
+                    if (_sqlServerDataStoreConfiguration.AllowDatabaseCreation)
                     {
-                        _logger.LogInformation("Database does not exist");
-
                         bool created = await CreateDatabaseAsync(connection, databaseName, cancellationToken);
 
                         if (created)
@@ -226,6 +226,12 @@ namespace Microsoft.Health.SqlServer.Features.Schema
                             _logger.LogWarning("Insufficient permissions to create the database");
                             return false;
                         }
+                    }
+                    else
+                    {
+                        _logger.LogWarning("Database creation is not enabled.");
+
+                        throw new InvalidOperationException(Resources.NewDatabaseCannotBeCreated);
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -232,6 +227,12 @@ private async Task<bool> CanInitializeAsync(CancellationToken cancellationToken)
                             return false;
                         }
                     }
+                    else
+                    {
+                        _logger.LogWarning("Database creation is not enabled.");
+
+                        throw new InvalidOperationException(Resources.NewDatabaseCannotBeCreated);
+                    }
```
rotodd, line 235: Before, if the database didn't exist and our config settings were set to the following:

`SqlServer:Initialize = true`
`SqlServer:AllowDatabaseCreation = false`

we would throw an error about incorrect login credentials and startup would fail. If there is no database and database creation is not enabled, it makes sense that startup would fail, but we should return a more useful error message. It wasn't that often that these two config settings were set to different values, so we never really ran into this.

Now that `SqlServer:Initialize` is removed and the error handling is updated, startup will fail if the database doesn't exist and `SqlServer:AllowDatabaseCreation = false`, and an appropriate message is returned.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -232,6 +227,12 @@ private async Task<bool> CanInitializeAsync(CancellationToken cancellationToken)
                             return false;
                         }
                     }
+                    else
+                    {
+                        _logger.LogWarning("Database creation is not enabled.");
+
+                        throw new InvalidOperationException(Resources.NewDatabaseCannotBeCreated);
```
rotodd, line 234: Is this the right type of exception to be throwing here?

test/Microsoft.Health.SqlServer.Web/Properties/launchSettings.json
```diff
@@ -4,9 +4,8 @@
       "commandName": "Project",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "development",
-        "SqlServer:Initialize": "false",
         "SqlServer:SchemaOptions:AutomaticUpdatesEnabled": "true",
-        "SqlServer:AllowDatabaseCreation": "false",
+        "SqlServer:AllowDatabaseCreation": "true",
```
rbans96, line 8: We might have to keep this to 'false' to not allow the E2E tests to try creating database until we add the ability to create the database in Github actions. I think E2E tests are currently failing on this PR due to this reason.

rotodd: Ah, thank you for looking into this. I originally set this to `true` to get the unit tests to pass. I'll figure out a different way around the problem!

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
rotodd: As discussed with @rbans96 and @brandonpollett, I removed the exception that is thrown and instead return false in 253f2af. An exception should be thrown once database creation in PR environments is enabled. I created a story for this [here](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77870).

test/Microsoft.Health.SqlServer.Web/Properties/launchSettings.json
rotodd: Reverted this to `false` in 253f2af.

rotodd:
It is difficult to merge a clean implementation of this until SQL database creation is enabled for this github repo. I will close this PR for now, and reopen a new one once [this](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77870) is done.

PR #130, opened 2020-11-30
Generate code for table types in separate files

johnstairs:
SQL table types, used for table-valued parameters, are immutable: there is no `ALTER` statement for them. For each table type definition that we encounter in .sql files, we generate a struct and a class, which are used for populating rows for table-valued parameters. 

When we want to change a stored procedure to take a new table type as a parameter, we generally can't do an `ALTER PROCEDURE` because that would break compatibility. So we create a new procedure with the new parameter signature. Currently, this requires duplicating all the row generators, because the types we generate are all nested within a `V<x>` class.

We would like to only need to implement a row generator for the new table type, which is the goal of this change. 

We now generate the CLR types for table types in a separate file, which are then referenced by stored procedure methods in the `VLatest` or `V<x>` classes. Since these types are immutable. they can safely be shared.


tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
```diff
@@ -207,13 +207,25 @@ protected static string GetRowStructNameForTableType(SchemaObjectName objectName
...
-        /// Strips away the version suffix from a table type name.
+        /// Strips away the underscore in a in a suffix for a table type name.
+        /// For example, MyTableType_1 will return MyTableType1
         /// </summary>
         /// <param name="objectName">The table type name</param>
         /// <returns>The name</returns>
         private static string GetTableTypeNameWithoutVersionSuffix(SchemaObjectName objectName)
         {
-            return Regex.Replace(objectName.BaseIdentifier.Value, @"_\d+", string.Empty);
+            return Regex.Replace(objectName.BaseIdentifier.Value, @"_(\d+)$", "$1");
```
brendankowitz, line 217: I can't remember if we wanted to have the V in there too, do you remember? `MyTableTypeV1`

test/Microsoft.Health.SqlServer.Web/Features/Schema/Model/VLatest.Generated.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Microsoft.Health.SqlServer.Web.Features.Schema.Model
+{
+    using Microsoft.Health.SqlServer.Features.Client;
+    using Microsoft.Health.SqlServer.Features.Schema.Model;
+
+    internal class VLatest
+    {
+        internal readonly static InsertNumbersProcedure InsertNumbers = new InsertNumbersProcedure();
+        internal readonly static MyProcedureProcedure MyProcedure = new MyProcedureProcedure();
+        internal class InsertNumbersProcedure : StoredProcedure
```
brendankowitz, line 18: Nit: +newline 😄

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
johnstairs: That's a good idea, I like it!

test/Microsoft.Health.SqlServer.Web/Features/Schema/Model/VLatest.Generated.cs
johnstairs: This is formatted using Roslyn's [`NormalizeWhitespace()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxnodeextensions.normalizewhitespace?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(Microsoft.CodeAnalysis.SyntaxNodeExtensions.NormalizeWhitespace%60%601);k(SolutionItemsProject);k(DevLang-csharp)%26rd%3Dtrue&view=roslyn-dotnet) method, which formats members like this. 
We can write/use another formatter, but I'd avoid including that change with this.

johnstairs: Actually, maybe a small change can fix this....

brendankowitz (approved):

test/Microsoft.Health.SqlServer.Web/Features/Schema/Migrations/2.sql
```diff
@@ -6,6 +6,32 @@
 
 INSERT INTO dbo.SchemaVersion
 VALUES
-    (2, 'started')
+    (3, 'started')
```
rbans96, line 9: Is it an intentional change to insert '3' in 2.sql or I might be missing something here.
I know it is little overhead and am sorry to not catch up on the user story to automate this as it got deprioritized on my board.

test/Microsoft.Health.SqlServer.Web/Features/Schema/Migrations/3.sql
```diff
@@ -0,0 +1,37 @@
+﻿-- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
+
+/*************************************************************
+    Schema Version
+**************************************************************/
+
+INSERT INTO dbo.SchemaVersion
+VALUES
+    (2, 'started')
```
rbans96, line 9: Same here. Probably (3, 'started')?

test/Microsoft.Health.SqlServer.Web/Features/Schema/Migrations/2.sql
johnstairs: No that was not intended :) thanks!

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
johnstairs: Fixed [here](https://github.com/microsoft/healthcare-shared-components/pull/130/commits/5fe612c6c12751cb1b1f0284e9e589adb86e950a)

test/Microsoft.Health.SqlServer.Web/Features/Schema/Model/VLatest.Generated.cs
johnstairs: Improved the overall formatting [here](https://github.com/microsoft/healthcare-shared-components/pull/130/commits/03534da7309fab973802c3275e7a34de65152edd)

PR #133, opened 2020-12-07
Adds support for .net5

brendankowitz:
## Description

Adds multi-target support for .net core 3.1 and .net 5
`<TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>`

## Semver Change
Feature (adds a supported runtime)


Directory.Build.props
```diff
@@ -11,6 +11,13 @@
     <!--This will target the latest patch release of the runtime released with the current SDK.  -->
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <RepositoryUrl>https://github.com/microsoft/healthcare-components</RepositoryUrl>
+
+    <SupportedFrameworks>netcoreapp3.1;net5.0</SupportedFrameworks>
+    <SdkPackageVersion>3.1.7</SdkPackageVersion>
```
Ivanidzo4ka, line 16: >3.1.7 [](start = 23, length = 5)

super nit: [https://versionsof.net/core/3.1/](https://versionsof.net/core/3.1/) latest one is `3.1.10` any reason to lock `3.1.7`?

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
johnstairs, line 4: What is the reason for not targeting net5.0 here?

Ivanidzo4ka, line 4: >1 [](start = 34, length = 1)

nit: Is this necessary?

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.nuspec
```diff
@@ -7,13 +7,14 @@
...
-        <repository type="git" url="https://github.com/microsoft/healthcare-components.git" />
+        <repository type="git" url="https://github.com/microsoft/healthcare-shared-components.git" />
     </metadata>
     <files>
         <file src="_._" target="lib/netstandard2.0/" />
         <file src="Microsoft.Health.Extensions.BuildTimeCodeGenerator.targets" target="build/netstandard2.0/Microsoft.Health.Extensions.BuildTimeCodeGenerator.targets" />
 
         <file src="Microsoft.Health.Extensions.BuildTimeCodeGenerator.targets" target="tools/" />
         <file src="$publishdir$\netcoreapp3.1\**\*" target="tools/netcoreapp3.1/" />
+        <file src="$publishdir$\net5.0\**\*" target="tools/net5.0/" />
```
Ivanidzo4ka, line 18: >file [](start = 9, length = 4)

it's gonna be one chunky nuget file :D

tools/SchemaManager/SchemaManager.csproj
```diff
@@ -39,4 +35,8 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Health.SqlServer\Microsoft.Health.SqlServer.csproj" />
```
Ivanidzo4ka, line 39: ><ProjectReference Include="..\..\src\Microsoft.Health.SqlServer\Microsoft.Health.SqlServer.csproj" /> [](start = 4, length = 101)

+1

src/Microsoft.Health.Client/Microsoft.Health.Client.csproj
```diff
@@ -1,13 +1,13 @@
...
-        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <TargetFrameworks>$(SupportedFrameworks);</TargetFrameworks>
     </PropertyGroup>
 
     <ItemGroup>
         <PackageReference Include="Ensure.That" Version="9.2.0" />
         <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.5.0" />
-        <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.7" />
-        <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="5.5.0" />    
+        <PackageReference Include="Microsoft.Extensions.Options" Version="$(SdkPackageVersion)" />
```
johnstairs, line 10: is this always the case?

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
Ivanidzo4ka: or $(SupportedFrameworks); as in other projects.

---
In reply to: [538783221](https://github.com/microsoft/healthcare-shared-components/pull/133#discussion_r538783221) [](ancestors = 538783221)

src/Microsoft.Health.Client/Microsoft.Health.Client.csproj
johnstairs: There are a few other instances of this. Not sure if it was deliberate.

Directory.Build.props
Ivanidzo4ka: I guess answer is to keep PR small and to not introduce extra changes, right?

---
In reply to: [538782058](https://github.com/microsoft/healthcare-shared-components/pull/133#discussion_r538782058) [](ancestors = 538782058)

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.nuspec
johnstairs: Yeah, I'm not sure this is quite the right design here. 50MB of runtime (now *2) doesn't seem right. 
Out of scope for this PR perhaps.

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
brendankowitz: netstandard supports more project types including full .net. I'd wait to upgrade this until downstream projects can unify platforms

brendankowitz: I think I'll revert this. In https://docs.microsoft.com/en-us/dotnet/standard/net-standard .net5 implements .net standard 2.1, .net 4.6.1 doesn't.
So we should keep 2.0 until we move it to .net5.

brendankowitz: This should be ok as standard until it needs to be updated:
> For existing code that targets netstandard, there's no need to change the TFM to net5.0. .NET 5.0 implements .NET Standard 2.1 and earlier. The only reason to retarget from .NET Standard to .NET 5.0 would be to gain access to more runtime features
https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-5-and-net-standard

Ivanidzo4ka: I'm just curious, does anyone consume Abstraction without other projects we have in this repo?

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.nuspec
brendankowitz: Yeah, this shouldn't be forever it should just give us a ramp to upgrade our projects and drop 3.1.
In size its ~30mb compressed as a nuget, even with the second version its still smaller than self-contained.

src/Microsoft.Health.Client/Microsoft.Health.Client.csproj
brendankowitz: Sorry just clarifying whats the case :P?
Most `Microsoft.Extensions.*` packages version together and are in sync with the runtime. I've added this variable to make that easier

Directory.Build.props
brendankowitz: That was my original intent, the only difference here should be the addition of .net5

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
brendankowitz: No, I wouldn't think so

Ivanidzo4ka (approved):

johnstairs (approved):

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.nuspec
johnstairs: I guess I'm surprised that a full publish is needed there.

PR #137, opened 2020-12-18
Set semver variables as output

LTA-Thinking:
## Description
Sets the semver variables as output so other steps in the build can use them.

## Related issues
Prevents semver from running in the linux build which isn't working currently


brendankowitz (approved):
:shipit:

PR #142, opened 2021-01-22
Increases SQL read column count

rotodd:
## Description
SQL reindexing requires an additional column in `dbo.Resource` for the search parameter hash value. This PR adds the ability to read an 11th column in the SQL search service in the FHIR server.

## Related issues
Addresses [AB#79096](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79096).

## Testing
Copied .pdb and .dll files and tested with the FHIR server locally.


feordin (approved):
Built and used new method in Fhir-server.  LGTM :).

PR #145, opened 2021-01-28
Fix property syntax

wsugarman:
## Description
- Fix build pipeline syntax to properly set build properties

## Related issues
N/A

## Testing
Validate ad-hoc pipeline run

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -17,7 +17,7 @@ steps:
   displayName: 'dotnet build $(buildConfiguration)'
   inputs:
     command: build
-    arguments: '--configuration $(buildConfiguration) /p:AssemblyVersion="$(assemblySemVer)" /p:FileVersion="$(assemblySemFileVer)" /p:InformationalVersion="$(informationalVersion)" /p:ContinuousIntegrationBuild=true /warnaserror'
+    arguments: '--configuration $(buildConfiguration) "-p:AssemblyVersion=$(assemblySemVer);FileVersion=$(assemblySemFileVer);InformationalVersion=$(informationalVersion);ContinuousIntegrationBuild=true"'
 
 - task: DotNetCoreCLI@2
   displayName: 'dotnet test UnitTests'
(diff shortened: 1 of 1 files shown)


build/build.yml
```diff
@@ -17,14 +17,14 @@ steps:
   displayName: 'dotnet build $(buildConfiguration)'
   inputs:
     command: build
-    arguments: '--configuration $(buildConfiguration) /p:AssemblyVersion="$(assemblySemVer)" /p:FileVersion="$(assemblySemFileVer)" /p:InformationalVersion="$(informationalVersion)" /p:ContinuousIntegrationBuild=true /warnaserror'
```
wsugarman, line 20: I don't think we need `/warnaserror` here, as the existing `<WarningsAsErrors/>` property in the *Directory.Build.props* seems to cover this as per the [documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2019)

build/build.yml
```diff
@@ -17,14 +17,14 @@ steps:
...
+    arguments: '--configuration $(buildConfiguration) -p:AssemblyVersion="$(assemblySemVer)" -p:FileVersion="$(assemblySemFileVer)" -p:InformationalVersion="$(informationalVersion)" -p:ContinuousIntegrationBuild=true'
 
 - task: DotNetCoreCLI@2
   displayName: 'dotnet test UnitTests'
   inputs:
     command: test
     projects: '**/*UnitTests/*.csproj'
-    arguments: '--configuration $(buildConfiguration)'
+    arguments: '--configuration $(buildConfiguration) --no-build'
```
wsugarman, line 27: This shouldn't build again, as we just built this configuration, but I'll be explicit as we wouldn't compile with the additional properties!

wsugarman:
It seems some of the assembly properties were not being set properly, as evidenced by inspecting the result of the CI build:
image

You can see here the `AssemblyFileVersion` and `ContinuousIntegrationBuild` settings were not set properly.
For those interested, this tool is the [NuGet Package Explorer](https://www.microsoft.com/store/productId/9WZDNCRDMDM3)

brandonpollett (approved):

PR #146, opened 2021-02-01
Adding support to read from and write to float column.

snehabandla:
Add support to handle float as a column's data type.  Per [documentation](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings), a SQL float maps to a .NET double.

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
@@ -217,6 +217,24 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Model
         }
     }
 
+    public class FloatColumn : Column<double>
+    {
+        public FloatColumn(string name, byte precision)
+            :base(name, SqlDbType.Float, precision, 0)
+        {
+        }
+
+        public override double Read(SqlDataReader reader, int ordinal)
+        {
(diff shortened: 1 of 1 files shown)


brandonpollett (approved):

src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -409,6 +427,31 @@ public override void Set(SqlDataRecord record, int ordinal, decimal? value)
...
+    public class NullableFloatColumn : Column<double?>
+    {
+        public NullableFloatColumn(string name, byte precision)
+            : base(name, SqlDbType.Float, true, precision, 0)
+        {
+        }
+
+        public override double? Read(SqlDataReader reader, int ordinal)
+        {
+            return reader.IsDBNull(Metadata.Name, ordinal) ? default(float?) : reader.GetDouble(Metadata.Name, ordinal);
```
johnstairs, line 439: should be `default(double?)`

snehabandla: fixed this!

johnstairs:
Looks good. For the generated code to compile, we'll need to specify the precision, which is probably the right thing to do anyway.

johnstairs (approved):

PR #149, opened 2021-02-11
Using correct SqlMetaData constructor for float column

snehabandla:
Reference:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.server.sqlmetadata.-ctor?view=dotnet-plat-ext-5.0
https://docs.microsoft.com/en-us/sql/t-sql/data-types/float-and-real-transact-sql?view=sql-server-ver15

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
@@ -442,7 +442,7 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Model
     public class NullableFloatColumn : Column<double?>
     {
         public NullableFloatColumn(string name, byte precision)
-            : base(name, SqlDbType.Float, true)
+            : base(name, SqlDbType.Float, true, 8, precision, 0, 0, SqlCompareOptions.None, null)
         {
         }
 
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Buffers;
 using System.Data;
+using System.Data.SqlTypes;
 using System.IO;
 using Microsoft.Data.SqlClient;
 using Microsoft.Data.SqlClient.Server;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -220,10 +232,26 @@ public override void Set(SqlDataRecord record, int ordinal, decimal value)
...
+            {
+                return 4;
+            }
+            else if (precision >= 24 && precision <= 53)
+            {
+                return 8;
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException(string.Format("Precision {0} must be between 1 & 53", precision));
```
brandonpollett, line 251: the error message should be in the resources file.

src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -220,10 +232,26 @@ public override void Set(SqlDataRecord record, int ordinal, decimal value)
     public class FloatColumn : Column<double>
     {
         public FloatColumn(string name, byte precision)
-            : base(name, SqlDbType.Float, false, precision, 0)
+            : base(name, SqlDbType.Float, false, GetLength(precision), precision, 0, 0, SqlCompareOptions.None, null)
         {
         }
 
+        private static long GetLength(byte precision)
```
brandonpollett, line 239: Any way to abstract this to where we're not copying it? Maybe a separate static class in this file?

brandonpollett (approved):

src/Microsoft.Health.SqlServer/Features/Schema/Model/ColumnUtilities.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+
+namespace Microsoft.Health.SqlServer.Features.Schema.Model
+{
+    internal static class ColumnUtilities
+    {
+        internal static long GetLengthForFloatColumn(byte precision)
+        {
+            if (precision >= 1 && precision <= 23)
```
johnstairs, line 14: In the docs, it looks like the upper bound is 24, not 23? https://docs.microsoft.com/en-us/sql/t-sql/data-types/float-and-real-transact-sql?view=sql-server-ver15#syntax

src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -220,7 +232,7 @@ public override void Set(SqlDataRecord record, int ordinal, decimal value)
     public class FloatColumn : Column<double>
     {
         public FloatColumn(string name, byte precision)
-            : base(name, SqlDbType.Float, false, precision, 0)
+            : base(name, SqlDbType.Float, false, ColumnUtilities.GetLengthForFloatColumn(precision), precision, 0, 0, SqlCompareOptions.None, null)
```
johnstairs, line 235: Kind of surprising, but it looks like this constructor, when called with float, throws ignores maxLength and precision. Bizarre api!

johnstairs: I mean the SqlMetadata contructor :)

johnstairs (approved):

PR #151, opened 2021-02-16
Adds RBAC related items to shared components

brandonpollett:
## Description
With both FHIR and DICOM having similar RBAC stories, adding the shared code to the shared components. Refactored to work on generic enums instead of specific `DataActions` so that they can be shared.

## Related issues
Addresses AB#79173.

## Testing
Role loading unit tests were added, manual integration testing on DICOM project.


src/Microsoft.Health.Core/Features/Context/GenericRequestContextAccessor.cs
```diff
@@ -0,0 +1,12 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Core.Features.Context
+{
+    public abstract class GenericRequestContextAccessor<T>
```
johnstairs, line 8: Maybe `RequestContextAccessor<T>`?

src/Microsoft.Health.Core/Features/Security/Role.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Globalization;
+using EnsureThat;
+
+namespace Microsoft.Health.Core.Features.Security
+{
+    /// <summary>
+    /// Class representing the concept of a role for the RBAC System
+    /// </summary>
+    /// <typeparam name="TEnum">Type representing the dataActions for the service</typeparam>
+    public class Role<TEnum>
```
johnstairs, line 16: `<TDataActions>` perhaps?

src/Microsoft.Health.Core/Features/Security/Authorization/RoleBasedAuthorizationService.cs
```diff
@@ -0,0 +1,62 @@
...
+            _rolesClaimName = authorizationConfiguration.RolesClaim;
+            _roles = authorizationConfiguration.Roles.ToDictionary(r => r.Name, StringComparer.OrdinalIgnoreCase);
+        }
+
+        public ValueTask<TEnum> CheckAccess(TEnum dataActions, CancellationToken cancellationToken)
+        {
+            ClaimsPrincipal principal = _genericRequestContextAccessor.RequestContext.Principal;
+
+            ulong permittedDataActions = 0;
+            ulong dataActionsUlong = Convert.ToUInt64(dataActions, NumberFormatInfo.InvariantInfo);
```
johnstairs, line 46: Instead of this approach, you can easily create a delegate that is compiled at runtime. Something like this:
``` csharp
private static readonly Func<TEnum, ulong> ConvertToULong = CreateConvertToULongFunc();

private static Func<TEnum, ulong> CreateConvertToULongFunc()
{
    var parameterExpression = Expression.Parameter(typeof(TEnum));
    return Expression.Lambda<Func<TEnum, UInt64>>(Expression.Convert(parameterExpression, typeof(UInt64)), parameterExpression).Compile();
}
```

... and another one for converting in the other direction

It should be significatly faster:
|          Method |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |                           |---------------- |-----------:|----------:|----------:|-------:|------:|------:|----------:|                           |    ConvertClass | 37.8900 ns | 0.7981 ns | 1.6302 ns | 0.0092 |     - |     - |      48 B |                           | ConvertDelegate |  0.9626 ns | 0.0574 ns | 0.0661 ns |      - |     - |     - |         - |

johnstairs (approved):

PR #153, opened 2021-02-18
Install .net 3.1 with global.json

brendankowitz:
## Description
Even though the netcore3.1/global.json doesn't have any project files, all global.json files are picked up by the `Use Dotnet` build task and installed. This should work for both external and internal builds.

Skip (build script)


KnicKnic (approved):

brendankowitz:
Added as part of #343

PR #156, opened 2021-02-22
Refactor to share tool code

rbans96:
## Description
Minor refactoring to execute paas schema by paas schema manager(https://microsofthealth.visualstudio.com/Health/_git/health-paas/pullrequest/14770).
Changes include -

- Updated SqlSchemaManager.applySchema as virtual method to override its behaviour in SchemaManager.Paas
- Updated assembly name Microsoft.Health -> Microsoft.Health.SchemaManager.Core to be consistent

## Related issues
Addresses [issue #].

## Testing
The commands are tested manually by running the SchemaManager tool

Patch|Skip|Feature|Breaking


src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -41,11 +41,6 @@ public static class SqlServerBaseRegistrationExtensions
                 .Singleton()
                 .AsSelf();
 
-            services.Add<SchemaManagerDataStore>()
```
rbans96, line 44: Removed redundant registration

tools/SchemaManager.Core/CommandOptions.cs
```diff
@@ -6,7 +6,7 @@
 using System;
 using System.CommandLine;
 
-namespace SchemaManager
+namespace SchemaManager.Core
```
YazanMSFT, line 9: My thinking when I refactored stuff out to SchemaManager.Core was that the common logic for running the tool would be in .Core, but the entry points would be in separate projects (for example, Command Line would be in the SchemaManager project, while the code for applying it on startup is in the SqlServer.Api project).

rbans96: Yeah, makes sense! Also, I was debating myself too to reuse this file or not. Because in paas, its not going to be used as tool rather job, and we don't need all of the options like next. Also we might need a different parameter instead of connection-string. So I think I should create a separate file for paas command options. Thoughts?

rbans96: Reverted the change in last commit!

YazanMSFT: Yup exactly, we'll have a separate file for those commands in that repo.

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -27,7 +27,10 @@ public SchemaClient()
 
         public void SetUri(Uri serverUrl)
         {
-            _httpClient.BaseAddress = serverUrl;
+            if (_httpClient.BaseAddress == null)
```
YazanMSFT, line 30: Why is this needed? If I were calling this method, I wouldn't expect that it can't be changed after the first call.

YazanMSFT (approved):

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
rbans96: Agree! But while I was doing some local testing, probably last week I noticed that if by chance SetUri is called while BaseAddress is already present, it throws exception. So I just wanted to ensure to set only if its null. Unfortunately, I can't repro this now and I don't remember exactly what were my local changes when I noticed this.

PR #157, opened 2021-02-26
Adds some integration tests for schema management

YazanMSFT:
## Description
Added an integration test project for Microsoft.Health.SqlServer. Currently this only runs on local host

## Related issues
Addresses [issue #].

## Testing
By running the tests


src/Microsoft.Health.SqlServer/Features/Schema/Manager/BaseSchemaRunner.cs
```diff
@@ -80,9 +80,17 @@ await Policy.Handle<SchemaManagerException>()
...
-            if (!await _schemaManagerDataStore.InstanceSchemaRecordExistsAsync(cancellationToken))
+            try
             {
-                throw new SchemaManagerException(Resources.InstanceSchemaRecordErrorMessage);
+                if (!await _schemaManagerDataStore.InstanceSchemaRecordExistsAsync(cancellationToken))
+                {
+                    throw new SchemaManagerException(Resources.InstanceSchemaRecordErrorMessage);
+                }
+            }
+            catch (SqlException e) when (e.Message.Contains("Invalid object name", StringComparison.OrdinalIgnoreCase))
```
rbans96, line 90: I am wondering when would this happen? Since before checking if record exists or not, we ensure that the base schema exists. So if it fails to create table as part of base schema execution, it should throw exception in _baseSchemaRunner.EnsureBaseSchemaExistsAsync.

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaInitializerTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        [Fact]
+        public async Task DoesDatabaseExist_DoesNotExist_ReturnsFalse()
+        {
+            Assert.False(await SchemaInitializer.DoesDatabaseExistAsync(Connection, "doesnotexist", CancellationToken.None));
+        }
+
+        [Fact]
+        public async Task DoesDataBaseExist_Exists_ReturnsTrue()
```
rbans96, line 28: nit : Should we use Given_When_Then for readability. 
E.g. GivenDatabaseExists_WhenCreateDatabaseAsync_ThenReturnsTrue

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaUpgradeRunnerTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        [Fact(Skip = "Sample schemas have version insertion which causes this to fail. Unignore when test scripts fixed.")]
+        public async Task ApplySchema_UpgradeSchema_Succeeds()
+        {
+            await _runner.ApplyBaseSchemaAsync(CancellationToken.None);
+            await _runner.ApplySchemaAsync(2, true, CancellationToken.None);
+            var version = await _schemaDataStore.GetCurrentSchemaVersionAsync(CancellationToken.None);
+            Assert.Equal(2, version);
+            await _runner.ApplySchemaAsync(2, false, CancellationToken.None);
```
rbans96, line 71: The version should be 3 instead of 2 here

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaUpgradeRunnerTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        [Fact]
+        public async Task ApplySchema_BaseSchemaExists_Succeeds()
+        {
+            await _runner.ApplyBaseSchemaAsync(CancellationToken.None);
+            await _runner.ApplySchemaAsync(1, true, CancellationToken.None);
+            var version = await _schemaDataStore.GetCurrentSchemaVersionAsync(CancellationToken.None);
+            Assert.Equal(1, version);
+        }
+
+        [Fact(Skip = "Sample schemas have version insertion which causes this to fail. Unignore when test scripts fixed.")]
```
rbans96, line 64: For my understanding, Is this 2.diff.sql, can we explicitly mention that, so that we know which file need to fix?

test/SchemaManager.Core.UnitTests/SqlSchemaManagerTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        [Fact]
+        public async Task ApplySchema_UsingDiffScript_Succeeds()
+        {
+            _schemaManagerDataStore.GetCurrentSchemaVersionAsync(default).ReturnsForAnyArgs(Task.FromResult(1));
+            _client.GetCurrentVersionInformationAsync(Arg.Any<CancellationToken>()).ReturnsForAnyArgs(new List<CurrentVersion> { });
+            _client.GetAvailabilityAsync(Arg.Any<CancellationToken>()).ReturnsForAnyArgs(new List<AvailableVersion> { new AvailableVersion(1, "_script/1.sql", "_script/1.diff.sql"), new AvailableVersion(2, "_script/2.sql", "_script/2.diff.sql") });
+            _client.GetCompatibilityAsync(Arg.Any<CancellationToken>()).ReturnsForAnyArgs(new CompatibleVersion(1, 2));
+            _client.GetDiffScriptAsync(Arg.Is<Uri>(new Uri("_script/2.diff.sql", UriKind.Relative)), Arg.Any<CancellationToken>()).Returns("script");
+            await _sqlSchemaManager.ApplySchema("connectionString", new Uri("https://localhost/"), new MutuallyExclusiveType { Latest = true, Version = 2, Next = true });
```
rbans96, line 101: nit: There is a command line validation that only one of the option[version, latest, next] should be set in the object MutuallyExclusiveType.
So I would rather create - new MutuallyExclusiveType { Latest = false, Version = 2, Next = false}). So that one should not get confused, that we can input all at the same time.

src/Microsoft.Health.SqlServer/Features/Schema/Manager/BaseSchemaRunner.cs
YazanMSFT: While that is true, EnsureInstanceSchemaRecordExistsAsync is a public method, so it's possible that the caller didn't initialize the base schema yet. Though I should make it have a different error message so it's clearer that the issue is the base schema wasn't initialized yet.

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaUpgradeRunnerTests.cs
YazanMSFT: Sure, I'll update the comment. Yeah it's in there, when I try to upgrade to instance 2 I get a unique index violation

YazanMSFT: Looks like the issue was what you caught below :)

test/SchemaManager.Core.UnitTests/SchemaManager.Core.UnitTests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
```
rbans96, line 5: I am wondering if we should have assemblyName for SchemaManager.Core and SchemaManager.Core.UnitTests to include Microsoft.Health?
<AssemblyName>Microsoft.Health.SchemaManager.Core</AssemblyName>
<AssemblyName>Microsoft.Health.SchemaManager.Core.UnitTests</AssemblyName>

test/SchemaManager.Core.UnitTests/SqlSchemaManagerTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        [Fact]
+        public async Task ApplySchema_UsingDiffScript_Succeeds()
+        {
+            _schemaManagerDataStore.GetCurrentSchemaVersionAsync(default).ReturnsForAnyArgs(Task.FromResult(1));
+            _client.GetCurrentVersionInformationAsync(Arg.Any<CancellationToken>()).ReturnsForAnyArgs(new List<CurrentVersion> { });
+            _client.GetAvailabilityAsync(Arg.Any<CancellationToken>()).ReturnsForAnyArgs(new List<AvailableVersion> { new AvailableVersion(1, "_script/1.sql", "_script/1.diff.sql"), new AvailableVersion(2, "_script/2.sql", "_script/2.diff.sql") });
+            _client.GetCompatibilityAsync(Arg.Any<CancellationToken>()).ReturnsForAnyArgs(new CompatibleVersion(1, 2));
+            _client.GetDiffScriptAsync(Arg.Is<Uri>(new Uri("_script/2.diff.sql", UriKind.Relative)), Arg.Any<CancellationToken>()).Returns("script");
+            await _sqlSchemaManager.ApplySchema("connectionString", new Uri("https://localhost/"), new MutuallyExclusiveType { Latest = false, Version = 2, Next = true });
```
rbans96, line 101: Can we set false to Next as well. Since Latest, Version and Next are mutually exclusive and command validation ensures that only of them is present. Probably : ew MutuallyExclusiveType { Latest = false, Version = 2, Next = true }) -> new MutuallyExclusiveType { Version = 2 }? Though version =2 and Next resolves to the same schema version here :)

rbans96 (approved):

test/SchemaManager.Core.UnitTests/SchemaManager.Core.UnitTests.csproj
YazanMSFT: It would make sense, I didn't add the prefix since SchemaManager didn't have it either. Should I add it to SchemaManager too? This might be considered a breaking change and require bumping up the major version number on the nuget package.

rbans96: SchemaManager has the assembly name - https://github.com/microsoft/healthcare-shared-components/blob/master/tools/SchemaManager/SchemaManager.csproj#L6

PR #185, opened 2021-03-10
Add uploadmaximumConcurrency

smithago:
## Description
Add the upload concurrency setting, now that this setting can be used in uploadAsync.
Reaction to issue fixed https://github.com/Azure/azure-sdk-for-net/issues/12445

## Related issues
[AB#80301](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80301)

## Testing
Just added a new configuration setting. Testing will be done in dicom-server, that consumes this setting

### DIFF
--- a/src/Microsoft.Health.Blob/Configs/BlobDataStoreRequestOptions.cs
+++ b/src/Microsoft.Health.Blob/Configs/BlobDataStoreRequestOptions.cs
@@ -13,6 +13,8 @@ namespace Microsoft.Health.Blob.Configs
 
         public int ServerTimeoutInMinutes { get; set; } = 2;
 
-        public int DownloadMaximumConcurrency { get; set; } = 1;
+        public int DownloadMaximumConcurrency { get; set; } = 5;
+
+        public int UploadMaximumConcurrency { get; set; } = 5;
     }
 }


src/Microsoft.Health.Blob/Configs/BlobDataStoreRequestOptions.cs
```diff
@@ -13,6 +13,8 @@ public class BlobDataStoreRequestOptions
 
         public int ServerTimeoutInMinutes { get; set; } = 2;
 
-        public int DownloadMaximumConcurrency { get; set; } = 1;
+        public int DownloadMaximumConcurrency { get; set; } = 5;
```
smithago, line 16: default value is 5 in SDK, here https://github.com/Azure/azure-sdk-for-net/blob/71d3750b48129b42b30d1a95ed9a906aba72648b/sdk/storage/Azure.Storage.Common/src/Shared/Constants.cs#L198

src/Microsoft.Health.Blob/Configs/BlobDataStoreRequestOptions.cs
```diff
@@ -13,6 +13,8 @@ public class BlobDataStoreRequestOptions
...
+
+        public int UploadMaximumConcurrency { get; set; } = 5;
```
brandonpollett, line 18: What is this setting used in? I don't see it the constants.

brandonpollett (approved):

PR #190, opened 2021-03-17
Use .NET Analyzers from .NET 5 SDK

wsugarman:
## Description
The .NET 5 SDK includes static code analyzers that replace the Roslyn analyzer NuGet packages. As part of this migration, I have:
- Included an *.editorconfig* file for style and rule configuration
- Addressed a number  of "low-hanging fruit" suggestions/warnings/errors from the analyzers. Common complaints from the analyzers include:
    - `null`-check arguments before use in `public` methods
    - Match method parameter names between interfaces and implementations
    - Attributes are typically marked as `sealed`
    - Use `ConfigureAwait(false)` in library code (even though ASP.NET Core does not have a custom synchronization context, it is good to be agnostic)
    - Assemblies should indicate their CLS compliance
    - `abstract` types should not have `public` constructors (they should be `protected`)
    - Normalize use of `var`; only use when the resulting type is "obvious"
- For more substantial changes, I will point them out in comments
- As part of the breaking change, I also corrected the various NuGet dependencies across the repository

## Related issues
N/A

## Testing
Unit Tests

Breaking


src/Microsoft.Health.Abstractions/Properties/AssemblyInfo.cs
```diff
@@ -3,6 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Resources;
 
+[assembly: CLSCompliant(false)]
```
wsugarman, line 9: Each of the assemblies is marked to indicate it does ensure compliance with the Common Language Specification (CLS)

.editorconfig
```diff
@@ -0,0 +1,186 @@
+﻿# editorconfig.org
```
wsugarman, line 1: I am re-using the file from DICOM

Directory.Build.props
```diff
@@ -9,18 +9,22 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <HighEntropyVA>true</HighEntropyVA>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
```
wsugarman, line 12: We'll use C# 9 explicitly, as we multi-target and by default the language is determined by the target framework

Directory.Build.props
```diff
@@ -9,18 +9,22 @@
...
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <Product>Microsoft Health</Product>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <RepositoryUrl>https://github.com/microsoft/healthcare-components</RepositoryUrl>
```
wsugarman, line 16: This is implied

Directory.Build.props
```diff
@@ -9,18 +9,22 @@
...
     <SdkPackageVersion>3.1.13</SdkPackageVersion>
     <SupportedFrameworks>netcoreapp3.1;net5.0</SupportedFrameworks>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors />
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- CA1031 - Allow Exception to be caught -->
+    <NoWarn>$(NoWarn);CA1031</NoWarn>
```
wsugarman, line 25: We often catch all exceptions and should probably re-evaluate its usage in some places

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -62,7 +62,7 @@ public string GetAuditEventType(string controllerName, string actionName)
             return null;
         }
 
-        Task IHostedService.StartAsync(CancellationToken cancellationToken)
+        public Task StartAsync(CancellationToken cancellationToken)
```
wsugarman, line 65: Here I wasn't sure why `Start` was explicitly implemented but `Stop` wasn't

src/Microsoft.Health.Api/Features/Audit/AuditLoggingFilterAttribute.cs
```diff
@@ -4,13 +4,16 @@
...
+using System.Diagnostics.CodeAnalysis;
 using EnsureThat;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.Health.Core.Features.Security;
 
 namespace Microsoft.Health.Api.Features.Audit
 {
     [AttributeUsage(AttributeTargets.Class)]
+    [SuppressMessage("Design", "CA1019:Define accessors for attribute arguments", Justification = "Derived classes will choose what to expose publicly.")]
+    [SuppressMessage("Performance", "CA1813:Avoid unsealed attributes", Justification = "This attribute to meant to be extended.")]
```
wsugarman, line 16: Whoops -- a typo!

src/Microsoft.Health.Blob/Features/Storage/BlobClientProvider.cs
```diff
@@ -42,7 +42,7 @@ public class BlobClientProvider : IHostedService, IRequireInitializationOnFirstR
         public Task StartAsync(CancellationToken cancellationToken)
         {
             // The result is ignored and will be awaited in EnsureInitialized(). Exceptions are logged within DocumentClientInitializer.
-            _ = _initializationOperation.EnsureInitialized();
+            _ = _initializationOperation.EnsureInitialized().AsTask();
```
wsugarman, line 45: `ValueTask` is not meant to run in this manner; it is meant to be immediately awaited and may cause issues with invoking `AsTask()`

src/Microsoft.Health.Blob/Features/Storage/BlobClientReadWriteTestProvider.cs
```diff
@@ -38,22 +38,20 @@ public async Task PerformTestAsync(BlobServiceClient client, BlobDataStoreConfig
...
             await blob.UploadAsync(
-                new MemoryStream(Encoding.UTF8.GetBytes(TestBlobContent)),
-                new BlobHttpHeaders()
-                {
-                    ContentType = "text/plain",
-                });
-            await DownloadBlobContentAsync(blob, cancellationToken);
+                content,
+                new BlobHttpHeaders { ContentType = "text/plain" },
+                cancellationToken: cancellationToken).ConfigureAwait(false);
```
wsugarman, line 45: Propagate the token

src/Microsoft.Health.Client/AuthenticationType.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.Health.Client
 {
     public enum AuthenticationType
     {
+        None = 0,
```
wsugarman, line 10: *Typically* enums have their default values defined

src/Microsoft.Health.Client/OAuth2ClientCredentialProvider.cs
```diff
@@ -46,9 +46,13 @@ protected override async Task<string> BearerTokenFunction(CancellationToken canc
             };
 
             using var formContent = new FormUrlEncodedContent(formData);
-            using HttpResponseMessage tokenResponse = await _httpClient.PostAsync(_oAuth2ClientCredentialConfiguration.TokenUri, formContent, cancellationToken);
+            using HttpResponseMessage tokenResponse = await _httpClient.PostAsync(_oAuth2ClientCredentialConfiguration.TokenUri, formContent, cancellationToken).ConfigureAwait(false);
 
-            var openIdConnectMessage = new OpenIdConnectMessage(await tokenResponse.Content.ReadAsStringAsync());
+#if NET5_0_OR_GREATER
```
wsugarman, line 51: Ugh.

Unfortunately only .NET 5 has the overload where we can propagate the token

src/Microsoft.Health.Core.UnitTests/Features/Security/RoleLoaderTests.cs
```diff
@@ -194,7 +194,7 @@ public async Task GivenValidDataActions_WhenSpecifiedAsRoleActions_AreRecognized
                     new
                     {
                         name = $"role{a}",
-                        dataActions = new[] { char.ToLower(a.ToString()[0]) + a.ToString().Substring(1) },
+                        dataActions = new[] { char.ToLower(a.ToString()[0]) + a.ToString()[1..] },
```
wsugarman, line 197: Ranges!

src/Microsoft.Health.Core/ActionTimer.cs
```diff
@@ -30,10 +30,7 @@ public ActionTimer(ILogger logger, string componentName)
             _stopwatch.Start();
         }
 
-        public TimeSpan GetElapsedTime()
-        {
-            return _stopwatch.Elapsed;
-        }
+        public TimeSpan ElapsedTime => _stopwatch.Elapsed;
```
wsugarman, line 33: This is a very breaking change, but I am also making other breaking changes here. So I thought I'd include it (the analyzer also flags this as something that should be a property)

src/Microsoft.Health.Core/Features/Security/RoleLoader.cs
```diff
@@ -33,20 +34,19 @@ public abstract class RoleLoader<TDataActions> : IHostedService
         where TDataActions : Enum
     {
         private readonly AuthorizationConfiguration<TDataActions> _authorizationConfiguration;
-        private readonly IHostEnvironment _hostEnvironment;
         private readonly Microsoft.Extensions.FileProviders.IFileProvider _fileProvider;
 
-        public RoleLoader(AuthorizationConfiguration<TDataActions> authorizationConfiguration, IHostEnvironment hostEnvironment)
+        protected RoleLoader(AuthorizationConfiguration<TDataActions> authorizationConfiguration, IHostEnvironment hostEnvironment)
```
wsugarman, line 39: An example of a ctor that should be `protected`

src/Microsoft.Health.Core/Features/Security/RoleLoader.cs
```diff
@@ -62,7 +62,7 @@ public Task StartAsync(CancellationToken cancellationToken)
             };
 
             validatingReader.ValidationEventHandler += (sender, args) =>
-                throw new InvalidDefinitionException(string.Format(Resources.ErrorValidatingRoles, args.Message));
+                throw new InvalidDefinitionException(string.Format(CultureInfo.InvariantCulture, Resources.ErrorValidatingRoles, args.Message));
```
wsugarman, line 65: Often our format string args are missing the explicit culture. For `string` args, the culture is irrelevant as the `string` is already a `string` and cannot be formatted for a given culture. `int` also doesn't vary much (if at all? I cannot think of anything). So I think Invariant is fine

src/Microsoft.Health.Core/RetryableInitializationOperation.cs
```diff
@@ -18,13 +19,20 @@ namespace Microsoft.Health.Core
...
+        private readonly bool _continueOnCapturedContext;
         private SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
         private Task _task;
 
         public RetryableInitializationOperation(Func<Task> operation)
+            : this(operation, false)
+        {
+        }
+
+        public RetryableInitializationOperation(Func<Task> operation, bool continueOnCapturedContext)
```
wsugarman, line 31: Because we don't make any assumptions about the task we're given, we should allow the user to specify the argument for `ConfigureAwait`. In fact, the lib Polly follows this pattern as well on its `ExecuteAsync` methods

src/Microsoft.Health.Extensions.DependencyInjection/Microsoft.Health.Extensions.DependencyInjection.csproj
```diff
@@ -5,6 +5,10 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
```
wsugarman, line 8: This was unintended and will be removed

src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -15,11 +16,13 @@
 using Microsoft.Health.SqlServer.Features.Routing;
 using Microsoft.Health.SqlServer.Features.Schema;
 using Microsoft.Health.SqlServer.Features.Schema.Extensions;
+using Microsoft.Health.SqlServer.Features.Schema.Messages.Get;
 
 namespace Microsoft.Health.SqlServer.Api.Controllers
 {
     [HttpExceptionFilter]
     [Route(KnownRoutes.SchemaRoot)]
+    [SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "ASP.NET Core does not have custom synchronization context.")]
```
wsugarman, line 25: I think it's probably reasonable to add `ConfigureAwait(false)` here too, as this is specifically for ASP.NET Core which does not include a custom synchronization context. Furthermore, it reinforces that the code here is agnostic to any specific scheduling needs

src/Microsoft.Health.SqlServer.UnitTests/IdentifierTests.cs
```diff
@@ -0,0 +1,73 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Xunit;
+
+namespace Microsoft.Health.SqlServer.UnitTests
+{
+    public class IdentifierTests
```
wsugarman, line 10: Tests for the database name validation

src/Microsoft.Health.SqlServer/Features/Schema/Manager/Model/CurrentVersion.cs
```diff
@@ -24,6 +24,6 @@ public CurrentVersion(int id, string status, List<string> servers)
 
         public string Status { get; }
 
-        public List<string> Servers { get; }
+        public IReadOnlyList<string> Servers { get; }
```
wsugarman, line 27: Exposing read-only yet mutable list seems counterintuitive

src/Microsoft.Health.SqlServer/Features/Schema/Model/TableValuedParameterDefinition.cs
```diff
@@ -29,7 +29,7 @@ protected TableValuedParameterDefinition(string parameterName, string tableTypeN
             _tableTypeName = tableTypeName;
         }
 
-        private SqlMetaData[] ColumnMetadata => _columnMetadata ?? (_columnMetadata = Columns.Select(c => c.Metadata).ToArray());
+        private SqlMetaData[] ColumnMetadata => _columnMetadata ??= Columns.Select(c => c.Metadata).ToArray();
```
wsugarman, line 32: This should be equivalent

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -87,38 +89,32 @@ public async Task InitializeAsync(bool forceIncrementalSchemaUpgrade = false, Ca
...
 
-                string tableName = "dbo.SchemaVersion";
+            const string tableName = "dbo.SchemaVersion";
 
-                // Since now the status is made consistent as 'completed', we might have to check for 'complete' as well for the previous version's status
-                using (var selectCommand = connection.CreateCommand())
-                {
-                    selectCommand.CommandText = string.Format(
-                        "SELECT MAX(Version) FROM {0} " +
-                        "WHERE Status = 'complete' OR Status = 'completed'", tableName);
```
wsugarman, line 111: The analyzer erroneously flagged this as a possible SQL injection because of the `string.Format`, but the `tableName` is actually constant

src/Microsoft.Health.SqlServer/Features/Schema/ScriptProvider.cs
```diff
@@ -19,47 +19,33 @@ public string GetMigrationScript(int version, bool applyFullSchemaSnapshot)
...
-                }
-
-                var scriptBytes = new byte[fileStream.Length];
-                await fileStream.ReadAsync(scriptBytes, 0, scriptBytes.Length, cancellationToken);
-                return scriptBytes;
+                throw new FileNotFoundException(Resources.ScriptNotFound);
             }
+
+            var scriptBytes = new byte[fileStream.Length];
+            await fileStream.ReadAsync(scriptBytes.AsMemory(0, scriptBytes.Length), cancellationToken).ConfigureAwait(false);
```
wsugarman, line 47: Use the new, apparently more efficient, `Memory<T>` type.

src/Microsoft.Health.SqlServer/Identifier.cs
```diff
@@ -0,0 +1,45 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Health.SqlServer
+{
+    internal static class Identifier
```
wsugarman, line 10: TODO: Leave comment about the source of this logic

wsugarman: https://docs.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver15

src/Microsoft.Health.SqlServer/Resources.resx
```diff
@@ -157,6 +157,10 @@
   <data name="InsufficientTablesPermissionsMessage" xml:space="preserve">
     <value>Insufficient permissions to create tables in the database.</value>
   </data>
+  <data name="InvalidDatabaseName" xml:space="preserve">
+    <value>The database name {0}  is an invalid SQL identifier</value>
```
wsugarman, line 161: TODO: Added too many spaces here

src/Microsoft.Health.SqlServer/RetryableInitializationOperation.cs
```diff
@@ -40,6 +48,7 @@ public RetryableInitializationOperation(Func<Task> operation)
         /// synchronization so only one running task will exist at a time.
         /// </summary>
         /// <returns>A task representing the completion of the initialization operation.</returns>
+        [SuppressMessage("Maintainability", "CA1508:Avoid dead conditional code", Justification = "Task reference may be updated by another thread.")]
```
wsugarman, line 51: Analyzer erroneously thinks checking the `_task` reference after the semaphore is unnecessary

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -167,22 +163,24 @@ public static async Task<bool> DoesDatabaseExistAsync(SqlConnection sqlConnectio
             return false;
         }
 
+        [SuppressMessage("Security", "CA2100:Review SQL queries for security vulnerabilities", Justification = "Database name is validated before use.")]
         public static async Task<bool> CreateDatabaseAsync(SqlConnection connection, string databaseName, CancellationToken cancellationToken)
         {
-            using (var canCreateDatabaseCommand = new SqlCommand("SELECT count(*) FROM fn_my_permissions (NULL, 'DATABASE') WHERE permission_name = 'CREATE DATABASE'", connection))
+            if (!Identifier.IsValidDatabase(databaseName))
```
wsugarman, line 169: This was the rare exception in the code where we cannot use a parameterized query, so to avoid any malicious `databaseName` values I validate it via a new helper method.

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -167,22 +163,24 @@ public static async Task<bool> DoesDatabaseExistAsync(SqlConnection sqlConnectio
...
-                    }
-                }
-                else
-                {
-                    return false;
-                }
+                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, Resources.InvalidDatabaseName, databaseName), nameof(databaseName));
+            }
+
+            using var canCreateDatabaseCommand = new SqlCommand("SELECT count(*) FROM fn_my_permissions (NULL, 'DATABASE') WHERE permission_name = 'CREATE DATABASE'", connection);
```
wsugarman, line 174: There are many instances where the indentation has changed due to the newer `using` statements

src/Microsoft.Health.Api/Modules/CorsModule.cs
```diff
@@ -32,7 +32,7 @@ public void Load(IServiceCollection services)
         {
             EnsureArg.IsNotNull(services, nameof(services));
 
-            var corsPolicyBuilder = new CorsPolicyBuilder()
+            CorsPolicyBuilder corsPolicyBuilder = new CorsPolicyBuilder()
```
brandonpollett, line 35: nit: (style) we generally leave it as var when the type is obvious.

src/Microsoft.Health.Core/Features/Security/RoleLoader.cs
brandonpollett: I don't think we localize any of our resources in this codebase now because it's all based on API calls that have no culture information passed. Because of this behavior, we should be defaulting to en-us from https://github.com/microsoft/healthcare-shared-components/blob/master/tools/SchemaManager.Core/AssemblyInfo.cs

src/Microsoft.Health.SqlServer.UnitTests/IdentifierTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        [InlineData("@@tempTable")]
+        [InlineData("[Missing Bracket")]
+        [InlineData("\"Missing Quote")]
+        [InlineData("[Unescaped]Delimiter]")]
+        [InlineData("\"Unescaped\"Delimiter\"")]
+        [InlineData("[]")]
+        [InlineData("\"\"")]
+        [InlineData("foo DROP DATABASE Production --")] // SQL Injection
+        [InlineData("𐊗𐊕𐊐𐊎𐊆𐊍𐊆")] // Lycian (SMP Unicode Characters)
+        [InlineData("😀")] // Emoticons
```
brandonpollett, line 28: 💯

src/Microsoft.Health.SqlServer.UnitTests/IndentedStringBuilderTests.cs
```diff
@@ -14,8 +14,9 @@ public class IndentedStringBuilderTests
         [Fact]
         public void GivenAnIndentedStringBuilder_WhenUsingIndentedScopes_KeepsTrackOfIndentation()
         {
-            var sb = new IndentedStringBuilder(new StringBuilder())
-                .AppendLine("class Foo").AppendLine("{");
+            IndentedStringBuilder sb = new IndentedStringBuilder(new StringBuilder())
```
brandonpollett, line 17: nit: var

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -17,94 +18,114 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Manager
...
-                _httpClient.BaseAddress = serverUrl;
+                _httpClient.BaseAddress = uri;
             }
         }
 
         public async Task<List<CurrentVersion>> GetCurrentVersionInformationAsync(CancellationToken cancellationToken)
         {
-            var currentUrl = Slash + KnownRoutes.SchemaRoot + Slash + KnownRoutes.Current;
-            var response = await _httpClient.GetAsync(RelativeUrl(currentUrl), cancellationToken);
+            const string currentUrl = Slash + KnownRoutes.SchemaRoot + Slash + KnownRoutes.Current;
```
brandonpollett, line 39: Field?

brandonpollett: Alternatively, why don't we just build these up in `KnownRoutes.cs`?

src/Microsoft.Health.SqlServer/Resources.resx
brandonpollett: nit: missing period.

src/Microsoft.Health.Core/Features/Security/RoleLoader.cs
wsugarman: Ah yes! So `InvariantCulture` is appropriate then (as would something like `CultureInfo.GetCultureInfo("en-us")` or `CultureInfo.GetCultureInfo("en")` ).

src/Microsoft.Health.Api/Modules/CorsModule.cs
wsugarman: Oh yes, I know why the analyzer flagged this. It is because we're using method-chaining. If we only called the ctor, it'd like `var`, but with the additional methods it got confused and assumes the type is no longer obvious (although it is).

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
wsugarman: Moved over to `KnownRoutes`

brandonpollett (approved):

PR #197, opened 2021-03-29
Makes RetrySqlCommandWrapper public

rotodd:
## Description
In PR #22, support for retrying SQL transient errors was added.

I am looking to use this in the implementation of reindexing for SQL in the[ fhir-server repo](https://github.com/Microsoft/fhir-server), to catch transient errors that occur when updating reindex job records.

## Related issues
Addresses [AB#80852](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80852).

## Testing
All tests pass as before.

Minor (adding functionality)

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Client/RetrySqlCommandWrapper.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Client/RetrySqlCommandWrapper.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Health.SqlServer.Features.Client
     /// <summary>
     /// A wrapper around <see cref="SqlCommand"/> to provide automatic retries for transient errors.
     /// </summary>
-    internal class RetrySqlCommandWrapper : SqlCommandWrapper
+    public class RetrySqlCommandWrapper : SqlCommandWrapper
     {
         private readonly SqlCommandWrapper _sqlCommandWrapper;
         private readonly IAsyncPolicy _retryPolicy;
--- a/src/Microsoft.Health.SqlServer/Features/Client/IPollyRetryLoggerFactory.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Client/IPollyRetryLoggerFactory.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Health.SqlServer.Features.Client
     /// <summary>
     /// Provides functionality for creating a logger for the <see cref="Polly"/> retry policy.
     /// </summary>
-    internal interface IPollyRetryLoggerFactory
+    public interface IPollyRetryLoggerFactory
     {
         /// <summary>
         /// Creates a logger.


src/Microsoft.Health.SqlServer/Features/Client/RetrySqlCommandWrapper.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Health.SqlServer.Features.Client
     /// <summary>
     /// A wrapper around <see cref="SqlCommand"/> to provide automatic retries for transient errors.
     /// </summary>
-    internal class RetrySqlCommandWrapper : SqlCommandWrapper
+    public class RetrySqlCommandWrapper : SqlCommandWrapper
```
wsugarman, line 18: Should we also expose `RetrySqlCommandWrapperFactory`?

wsugarman, line 18: This has nothing to do with your code, but I noticed that the `ExecuteAsync` methods below capture the `cancellationToken` in their closures. As an optimization, we can avoid the (very slight) overhead by using the other `ExecuteAsync` overload that accepts a token directly.

Eg.
```csharp
_retryPolicy.ExecuteAsync(t => _sqlCommandWrapper.ExecuteNonQueryAsync(t), cancellationToken);
```

Furthermore, this will help avoid creating an unnecessary `Task` if the token is already cancelled by the time this method is called. Polly does something like:

```csharp
cancellationToken.ThrowIfCancellationRequested();
TResult result = await action(cancellationToken);
```

wsugarman (approved):

wsugarman:
This would typically be a *minor* version change as we're adding functionality

rotodd:
> This would typically be a _minor_ version change as we're adding functionality

Noted, thanks for the correction! I updated the PR description accordingly.

src/Microsoft.Health.SqlServer/Features/Client/RetrySqlCommandWrapper.cs
rotodd: Good question. In the fhir-server, we currently have:

```
services.Add<RetrySqlCommandWrapperFactory>()
                .Singleton()
                .AsSelf()
                .AsService<SqlCommandWrapperFactory>();
```

So we are already using retry SQL logic under the hood. The reason why we are exposing the `RetrySqlCommandWrapper` here is because, in some cases, it is useful to be able to create a custom retry policy to handle certain exceptions that are not marked as transient:

`var retrySqlCommandWrapper = new RetrySqlCommandWrapper(previousSqlCommandWrapper, customRetryPolicy);`

All this said there could definitely be a better way to go about this. I think it might make sense to merge this PR, put up the corresponding fhir-server PR to give a bit more context, and iterate (in either repo) if needed!

rotodd: Cool, thanks for pointing this out. This is tempting to fix now, but I think it belongs in its own PR where we could check all `ExecuteAsync` calls, so I created [AB#81043](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81043).

rotodd: [Here](https://github.com/microsoft/fhir-server/pull/1811 )is the link to the corresponding fhir-server PR

PR #214, opened 2021-04-13
Adds codegen for `CREATE VIEW` statements

johnstairs:
## Description
Support generating code for `CREATE VIEW` statements, like we do today for `CREATE TABLE` statements. View declarations require a little more work to determine the schema. We need to walk the definition and figure out the type of each column in the select statement based on the existing tables.

## Testing
I added a sample view to the sample web project.

Patch

### DIFF
--- a/tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateTableVisitor.cs
+++ b/tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateTableVisitor.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Health.Extensions.BuildTimeCodeGenerator.Sql
 
             // find the class we generated for the table
             var tableDeclaration = (ClassDeclarationSyntax)MembersToAdd.FirstOrDefault(m => m is ClassDeclarationSyntax c && c.Identifier.ValueText == classNameForTable)
-                                   ?? throw new InvalidOperationException($"table not found");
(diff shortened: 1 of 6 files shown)


tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateTableVisitor.cs
```diff
@@ -59,6 +39,125 @@ public override void Visit(CreateTableStatement node)
...
+            string tableAliasName = columnReference.MultiPartIdentifier[0].Value;
+            string tableColumnName = columnReference.MultiPartIdentifier[1].Value;
+
+            string tableName = tablesInScope.Where(t => t.alias == tableAliasName).Select(t => t.name).FirstOrDefault() ?? throw new InvalidOperationException($"Unable to resolve table alias '{tableAliasName}'.");
+
+            string classNameForTable = GetClassNameForTable(tableName);
+
+            // find the class we generated for the table
+            var tableDeclaration = (ClassDeclarationSyntax)MembersToAdd.FirstOrDefault(m => m is ClassDeclarationSyntax c && c.Identifier.ValueText == classNameForTable)
+                                   ?? throw new InvalidOperationException($"table not found");
```
brendankowitz, line 137: Nit: Perhaps specify which table

```suggestion
                                   ?? throw new InvalidOperationException($"Table '{classNameForTable}' was not found");
```

brendankowitz (approved):

PR #220, opened 2021-04-24
Execute idempotent schema upgrades without transaction

rbans96:
## Description
This PR updates the Sql schema upgrade scripts to execute without transaction. Because the transactions are blocking and the schema upgrade script are now idempotent, we can get the improved performance with this change. And if the schema upgrade script is failed halfway through, we will be able to re-run it.

## Related issues
Addresses [#AB81399](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=81399)

## Testing
Unit tests are added and tested manually.

Patch|Skip|Feature|Breaking

### DIFF
--- a/tools/SchemaManager.Core/SqlSchemaManager.cs
+++ b/tools/SchemaManager.Core/SqlSchemaManager.cs
@@ -142,7 +142,12 @@ namespace SchemaManager.Core
 
                     string script = await GetScriptAsync(executingVersion, availableVersion.ScriptUri, token, availableVersion.DiffUri).ConfigureAwait(false);
 
-                    await UpgradeSchemaAsync(executingVersion, script, token).ConfigureAwait(false);
+                    // check if the record for given version exists in failed status
+                    await _schemaManagerDataStore.DeleteSchemaVersionAsync(executingVersion, SchemaVersionStatus.Failed.ToString(), token).ConfigureAwait(false);
+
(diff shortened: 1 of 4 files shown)


tools/SchemaManager.Core/SqlSchemaManager.cs
```diff
@@ -267,10 +272,10 @@ private async Task<string> GetScriptAsync(int version, string scriptUri, Cancell
 
         private async Task UpgradeSchemaAsync(int version, string script, CancellationToken cancellationToken)
```
YazanMSFT, line 273: Looks like this is specifically for initializing the schema now, can you rename it so that's clear?

tools/SchemaManager.Core/SqlSchemaManager.cs
```diff
@@ -142,7 +142,12 @@ await Policy.Handle<SchemaManagerException>()
 
                     string script = await GetScriptAsync(executingVersion, availableVersion.ScriptUri, token, availableVersion.DiffUri).ConfigureAwait(false);
 
-                    await UpgradeSchemaAsync(executingVersion, script, token).ConfigureAwait(false);
+                    // check if the record for given version exists in failed status
```
YazanMSFT, line 145: It looks like we use the diff scripts when availableVersions doesn't start with "1". I think the FHIR service doesn't include 1 and 2 anymore, does that mean we'll always initialize it using the diff scripts?

YazanMSFT (approved):

PR #221, opened 2021-04-26
Reverts making RetrySqlCommandWrapper and IPollyRetryLoggerFactory public

rotodd:
## Description
This PR reverts the changes made in #197, since we no longer need retry logic to be accessible in the FHIR server (see closing comment [here](https://github.com/microsoft/fhir-server/pull/1811#issuecomment-821733971)).

## Related issues
Addresses [AB#81418](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81418).

## Testing
All tests pass as before.

Patch

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Client/IPollyRetryLoggerFactory.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Client/IPollyRetryLoggerFactory.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Health.SqlServer.Features.Client
     /// <summary>
     /// Provides functionality for creating a logger for the <see cref="Polly"/> retry policy.
     /// </summary>
-    public interface IPollyRetryLoggerFactory
+    internal interface IPollyRetryLoggerFactory
     {
         /// <summary>
         /// Creates a logger.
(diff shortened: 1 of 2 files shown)


wsugarman (approved):

wsugarman (approved):
This is technically a breaking change 😟

But given we haven't used it... perhaps we can make this a patch and move on? Perhaps even unlist the previous version?

rotodd:
> This is technically a breaking change 😟
> 
> But given we haven't used it... perhaps we can make this a patch and move on? Perhaps even unlist the previous version?

@wsugarman good point :/. Happy to make this a patch. What is the best way to unlist the previous version?

wsugarman:
We would determine which version(s) introduced the addition of these classes as `public`, navigate to their respective [NuGet Feed](https://microsofthealthoss.visualstudio.com/FhirServer/_packaging?_a=package&feed=Public%40Local&package=Microsoft.Health.Dicom.SqlServer&protocolType=NuGet&version=3.0.44&view=versions), and get an admin for it to "unlist" those versions. They'll still be available to pull if someone already takes a dependency, but they won't be seen when browsing the package versions.

PR #233, opened 2021-05-21
Not to wrap transaction explicitly for schema initialization scripts

rbans96:
## Description
Since SQLServer schema initialization script could have non transactional script(i.e. which can't run in a transaction), so we don't want schema-manager to wrap initialization scripts in a transaction explicitly.
Instead, we would update the latest initialization scripts(x.sql) to include transaction in itself.

## Related issues
Addresses [#82137](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=82137)

## Testing
Unit tests are updated

Breaking

### DIFF
--- a/tools/SchemaManager.Core/SqlSchemaManager.cs
+++ b/tools/SchemaManager.Core/SqlSchemaManager.cs
@@ -117,7 +117,7 @@ namespace SchemaManager.Core
                     _logger.LogInformation(string.Format(CultureInfo.InvariantCulture, Resources.SchemaMigrationStartedMessage, availableVersions.Last().Id));
 
                     string script = await GetScriptAsync(1, availableVersions.Last().ScriptUri, token).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


tools/SchemaManager.Core/SqlSchemaManager.cs
```diff
@@ -275,12 +270,12 @@ private async Task<string> GetScriptAsync(int version, string scriptUri, Cancell
             return await _schemaClient.GetDiffScriptAsync(new Uri(diffUri, UriKind.Relative), cancellationToken).ConfigureAwait(false);
         }
 
-        private async Task InitializeSchemaAsync(int version, string script, CancellationToken cancellationToken)
+        private async Task UpgradeSchemaAsync(int version, string script, CancellationToken cancellationToken)
```
YazanMSFT, line 273: Maybe name this ApplySchemaInternalAsync? Since it's used for initialize too.

YazanMSFT (approved):

PR #240, opened 2021-06-02
SqlServer schema migration scripts guidelines

rbans96:
## Description
This PR adds the initial guidelines for SQLServer schema migration scripts

## Related issues
Addresses [issue #].

## Testing
NA

Patch|Skip|Feature|Breaking


brandonpollett (commented):
I think we need to add some guidance on a few more items:

- New non-nullable fields (default value, difference in new tables vs existing tables
- Updating procedure names when a new version is created (when you need to do this vs not)
- Appropriate use of `If not exists` for various types of sql objects.

Also, I think we can just say that if needed multiple internal transactions can be used if particular updates must either succeed or fail together and maybe give a concrete example of when you'd want to do this.

rbans96:
> I think we need to add some guidance on a few more items:
> 
> * New non-nullable fields (default value, difference in new tables vs existing tables
> * Updating procedure names when a new version is created (when you need to do this vs not)
> * Appropriate use of `If not exists` for various types of sql objects.
> 
> Also, I think we can just say that if needed multiple internal transactions can be used if particular updates must either succeed or fail together and maybe give a concrete example of when you'd want to do this.

Thanks Brandon, addressed new items.

brandonpollett (approved):

PR #248, opened 2021-06-10
Make audit event map distinct

jnlycklama:
## Description
- Audit event map to use distinct to get rid of duplicates. Duplicates will be introduced by API versioning since one event (ex: (Store, Post) => store audit type) will have two instances due to having two routes ("/v1.0/studies" & "/studies")
- Upgrade .net sdk to 5.0.301

## Related issues
Addresses [issue #[82754](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82754)].

## Testing
- Added unit test
- Created local nuget package, used with dicom-server to test it with the API version changes

Patch|Skip|Feature|Breaking


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "5.0.201"
+    "version": "5.0.301"
```
wsugarman, line 3: 👍

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -74,6 +74,7 @@ public Task StartAsync(CancellationToken cancellationToken)
                     return (ad.ControllerName, ad.ActionName, Attribute: attribute);
                 })
                 .Where(item => item.Attribute != null)
+                .Distinct()
```
wsugarman, line 77: I think there may be some strange cases where this doesn't work as expected. In particular, I am unsure what the value for `ActionName` will be if a controller uses multiple overloads of the same method name for its actions:

eg.
```csharp
public class ExampleController : Controller
{
    [Route("/Examples/{number}")]
    [AuditEventTypeAttribute("Get-By-Number")]
    public async Task GetAsync(int number) { ... }

    [Route("/Examples/{dt}")]
    [AuditEventTypeAttribute("Get-By-DateTime")]
    public async Task GetAsync(DateTime dt) { ... }
}
```

If they're the same, then I could image their respective tuples will not be filtered by `Distinct()` and could cause the same exception (presumably `ArgumentException`?). So, I think we need to push the de-duplication up a bit and perhaps throw an error when we get different attributes for the same `(ControllerName, ActionName)` tuple. That way we prevent unexpected behavior.

Of course, if this scenario is not possible then ignore me lol.

jnlycklama: Hm, good point! I just double-checked and you are right, Distinct() will not remove duplicate keys with different values. 

The existing code will throw an ArgumentException in this case. I guess we would want a more descriptive error message instead, so people know to pick a unique method name when they run into this error?

jnlycklama: Updated to throw a specific error for this case

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -74,9 +74,20 @@ public Task StartAsync(CancellationToken cancellationToken)
...
+                .GroupBy(item => (item.ControllerName, item.ActionName))
+                .Select(item =>
```
wsugarman, line 79: I think you can then merge this into the `GroupBy` and modify the group for easier use below, like:

```csharp
.GroupBy(
    x => (x.ControllerName, x.ActionName),
    x => x.Attribute,
    (key, values) =>
    {
        List<Attribute> attributes = values.ToList();
        if (attributes.Count > 1)
        {
            throw new DuplicateActionForAuditEventException(item.Key.ControllerName, item.Key.ActionName);
        }

        return (key, attributes[0]);
    }

File: src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs [hunk 4]
```diff
@@ -74,9 +74,20 @@ public Task StartAsync(CancellationToken cancellationToken)
...
+                {
+                    if (item.Count() != 1)
+                    {
+                        throw new DuplicateActionForAuditEventException(item.Key.ControllerName, item.Key.ActionName);
```
wsugarman, line 83: I appreciate the unique exception. You are also welcome to re-use exceptions where appropriate like perhaps `InvalidOperationException`. I mention this because (1) having too many very specific exceptions may be overwhelming to users and (2) creating exceptions can be kind of a pain as they should expose a [few standard ctors](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions), or even make them serializable as a best practice.

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
wsugarman: I began writing comments to explain that we're not handling the case where a developer annotates different actions with the same `AuditEventType`... but that scenario sounds out-of-scope at the moment. Perhaps just a `TODO` will suffice lol

wsugarman (approved):

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
jnlycklama: Ah makes sense. I've kept in this exception since it's already there, but updated it to have the correct constructors

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
jnlycklama: Thanks! Updated

src/Microsoft.Health.Api/Resources.resx
```diff
@@ -121,6 +121,10 @@
     <value>The maximum length of a custom audit header value is {0}. The supplied custom audit header '{1}' has length of {2}.</value>
     <comment>PH0 is a constant defining the max length of a header.  PH1 is the name  of the header sent in the request.  PH2 is the length of the incoming header value.  Example: The maximum length of a custom audit header value is 2048. The supplied custom audit header 'X-MS-AZUREHEALTH-AUDIT-SITE' has length of 3072.</comment>
   </data>
+  <data name="DuplicateActionForAuditEvent" xml:space="preserve">
+    <value>A duplicate controller/action was found with different audit events. Controller: {0}, Action: {1}. Try using distinct action names inside the same controller. </value>
```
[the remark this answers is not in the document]
jnlycklama: Fixed

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
jnlycklama: Added a TODO

PR #252, opened 2021-06-15
Add abstractions and common implementations needed to support Events.

kannanMSFT:
## Description

Add abstractions and common implementations needed to support Events.

## Testing

Have added unit tests.

Patch|Skip|Feature|Breaking


src/Microsoft.Health.Abstractions/Data/ISource.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Abstractions.Data
+{
+    /// <summary>
+    /// Used as the source for Data of type T.
+    /// </summary>
+    /// <typeparam name="T">Records of Type T</typeparam>
+    public interface ISource<T>
```
ms-teli, line 16: Same as above, we could be less general and more specific here. Maybe IEventSource ?

src/Microsoft.Health.Abstractions/Data/ISink.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Abstractions.Data
+{
+    /// <summary>
+    /// Generic Sink for writing data to any destination.
+    /// </summary>
+    /// <typeparam name="T">Generic data type T to write</typeparam>
+    public interface ISink<in T>
```
ms-teli, line 15: As being mentioned in the meeting, let's rename these? like IEventSink?

src/Microsoft.Health.Abstractions/Features/Events/IEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public interface IEvent
+    {
+        /// <summary>
+        /// Gets or sets the resource path of the event source. This must be set when publishing
+        ///     the event to a domain, and must not be set when publishing the event to a topic.
+        /// </summary>
+        public string Topic { get; set; }
+
+        /// <summary>Gets or sets a resource path relative to the topic path.</summary>
```
ms-teli, line 21: Nit, start a new line after<summary> to keep the format consistent?

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace Microsoft.Health.Core.Features.Events
+{
+    /// <summary>
+    /// EventGridSink.
+    /// </summary>
+    /// <typeparam name="T">T of Type IEvent.</typeparam>
+    public class EventGridSink<T> : ISink<T>
+        where T : class, IEvent
+    {
+        private readonly IEventGridPublisher _client;
```
ms-teli, line 23: Nit: should the variable name be _eventGridPublisher ?

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        /// <inheritdoc />
+        public async Task WriteAsync(T data)
+        {
+            EnsureArg.IsNotNull(data, nameof(data));
+
+            var eventGridEvent = new EventGridEvent(data.Subject, data.EventType, data.DataVersion, data.Data)
+            {
+                Topic = data.Topic, EventTime = data.EventTime, Id = data.Id,
```
ms-teli, line 52: Nit: multi-line seems better for reading.

src/Microsoft.Health.Abstractions/Features/Events/IEventGridPublisher.cs
```diff
@@ -0,0 +1,39 @@
...
+    public interface IEventGridPublisher
+    {
+        /// <summary>
+        /// SendEventAsync
+        /// </summary>
+        /// <param name="eventGridEvent">EventGridEvent</param>
+        /// <param name="cancellationToken">CancellationToken</param>
+        /// <returns>Task</returns>
+        public Task<Response> SendEventAsync(
+            EventGridEvent eventGridEvent,
```
ms-teli, line 26: Is "EventGridEvent" included in this PR?

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+           await _client.SendEventsAsync(eventGridEvents).ConfigureAwait(false);
+        }
+
+        /// <inheritdoc />
+        public async Task WriteAsync(T data)
+        {
+            EnsureArg.IsNotNull(data, nameof(data));
+
+            var eventGridEvent = new EventGridEvent(data.Subject, data.EventType, data.DataVersion, data.Data)
```
ms-teli, line 50: Why do we do an additional data transformation? eventGridEvent seems identical to data.

To securely get rid of it, can we do below? Assuming EventGridEvent implements IEvent
```
    public class EventGridSink : ISink<EventGridEvent>
```

I assume that EventGridSink should only be responsible for processing EventGridEvent model. It would be caller's job to make sure that model is passed in.

src/Microsoft.Health.Abstractions/Features/Events/IEventGridPublisher.cs
kannanMSFT: This EventGridEvent is from Azure lib

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
ms-teli: if the eventGridEvent is from Azure lib, then I think this makes more sense in your original version.

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
kannanMSFT: Yes, I should have renamed this also after refactor. Shall do.

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
kannanMSFT: EventGridEvent in Azure doesn't have an interface and hence abstracted it. Let me think about this from testing prespective.

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
kannanMSFT: fixed.

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
kannanMSFT: Fixed.

src/Microsoft.Health.Abstractions/Data/ISink.cs
kannanMSFT: My idea was this is a shared component at Microsoft.Health level and not related to any particular domain (FHIR or Dicom) and thus allows any one of the products to use them. But renaming them under .Data would be misleading.

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
kannanMSFT: Fixed. Good point. I should have revisited this after the POC.

src/Microsoft.Health.Abstractions/Data/ISink.cs
kannanMSFT: Fixed with rename to IChangeFeedSource<T> and IChangeFeedSink<T>

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        private async Task SendEventAsync(EventGridEvent eventGridEvent)
+        {
+            await _eventGridPublisher.SendEventAsync(eventGridEvent).ConfigureAwait(false);
+        }
+
+        private async Task SendEventsAsync(IEnumerable<EventGridEvent> eventGridEvents)
+        {
+           await _eventGridPublisher.SendEventsAsync(eventGridEvents).ConfigureAwait(false);
```
ms-teli, line 39: Nit: indention seems not right.

ms-teli (approved):

ms-teli:
Looks good to me. Please have someone from FHIR team taking a look before merging.

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
```diff
@@ -5,6 +5,11 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.3.0" />
```
brendankowitz, line 9: I don't think EventGrid should be added to Abstractions...

kannanMSFT: Originally it wasn't dependent but since we didn't want to have IEvent based on earlier comment here by Teng, I had to get rid of it and that introduces the dependency on EventGrid. Let me know.

kannanMSFT: Let me double check.

brendankowitz: This will introduce EventGrid as a dependency on all downstream projects that depend on Abstractions, I don't think this would be idea. Is there going to be a `Microsoft.Health.EventGrid` common library also?

kannanMSFT: Okay I remember now. Originally there was no dependency because I had abstracted EventGridEvent with IEvent. But since Teng wanted to get rid of that in earlier comments above, it introduced the dependency for IEventGridPublisher under Microsoft.Health.Abstractions.Features.Events because instead of taking IEvent its now taking EventGridEvent.

kannanMSFT: One fix would be to move the IEventGridPublisher to  Microsoft.Health.Core.Features.Events

kannanMSFT: Thanks for catching this @brendankowitz . I have moved it out of abstractions project, given we already got rid of IEvent. Fixed.

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
```diff
@@ -5,6 +5,9 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
```
brandonpollett, line 8: nit: not needed.

src/Microsoft.Health.Core/Microsoft.Health.Core.csproj
```diff
@@ -6,6 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.3.0" />
```
brandonpollett, line 9: I don't think this belongs here either. To follow the common patterns we have set, I would expect there to be a `Microsoft.Health.EventGrid` project that would hold `EventGrid` implementations of the various parts you're trying to implement.

src/Microsoft.Health.Core/Features/Events/IEventGridPublisher.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Messaging.EventGrid;
+
+namespace Microsoft.Health.Abstractions.Features.Events
```
brandonpollett, line 12: Generally would need to be in the same namespace as the file.

src/Microsoft.Health.Core/Features/Events/EventGridSink.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            await _eventGridPublisher.SendEventsAsync(eventGridEvents).ConfigureAwait(false);
+        }
+
+        /// <inheritdoc />
+        public async Task WriteAsync(EventGridEvent data)
+        {
+            EnsureArg.IsNotNull(data, nameof(data));
+
+            await SendEventAsync(data).ConfigureAwait(false);
```
brandonpollett, line 47: Any particular reason we're calling a private method here that has a single line? Why not just call `await _eventGridPublisher.SendEventAsync(eventGridEvent).ConfigureAwait(false);` here?

src/Microsoft.Health.Core/Features/Events/EventGridPublisher.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class EventGridPublisher : IEventGridPublisher
+    {
+        private readonly EventGridPublisherClient _client;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EventGridPublisher"/> class.
+        /// </summary>
+        /// <param name="endpoint">Uri of topic</param>
+        /// <param name="key">access key</param>
+        public EventGridPublisher(Uri endpoint, string key)
```
brandonpollett, line 31: What's the intended lifetime of this? Is it going to be a `singleton` or `scoped` or used by a factory? The reason I ask is it may inform how we want to handle this what I presume is a secret that will need to be rotated at some point.

src/Microsoft.Health.Abstractions/Features/Operations/IWorker.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Abstractions.Features.Operations
+{
+    /// <summary>
+    /// Abstraction for a worker
+    /// </summary>
+    public interface IWorker
```
brandonpollett, line 14: What's the point of this?

brandonpollett (changes requested):

src/Microsoft.Health.Core/Features/Events/IEventGridPublisher.cs
kannanMSFT: My bad. After the file move forgot to change it.

src/Microsoft.Health.Abstractions/Features/Operations/IWorker.cs
kannanMSFT: It was intended to be used later by IPublishEvetntsWorker and hopefully other workers. But after re-checking based on your comment the usages in other workers realized I can get rid of this and instead move it to IPublishEvetntsWorker since each worker seems to have a different signature.

src/Microsoft.Health.Core/Features/Events/EventGridPublisher.cs
kannanMSFT: Yes it would be a singleton injected later. We don't need this access key option now, given the PaaS would only use the client certificate option. But I left this in here so if someone wants to use on the OSS side with just the access key they still could.

src/Microsoft.Health.Core/Microsoft.Health.Core.csproj
kannanMSFT: I see. Okay. Let me add it there then.

src/Microsoft.Health.Abstractions/Data/IChangeFeedSink.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Write data to sink.
+        /// </summary>
+        /// <param name="data">Data of type T.</param>
+        /// <returns>Task.</returns>
+        Task WriteAsync(T data);
+
+        /// <summary>
+        /// WRite a collection of type T.
```
brandonpollett, line 25: nit: "Write"

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
+
```
brandonpollett, line 8: nit: extra line.

brandonpollett (approved):

PR #254, opened 2021-06-18
Library files to support checkpointing

wi-y:
## Description
This is a new feature to support storing checkpoints (watermarks) in blob storage, or potentially other locations in the future. This PR contains the minimal interfaces and classes needed to do this. It also includes a base class to upload and read checkpoints from blob.

## Testing
Defined interfaces and verified they will meet minimum use case when injected at startup.
Verified we can store and read checkpoint from blob.

Feature

### DIFF
--- a/Microsoft.Health.sln
+++ b/Microsoft.Health.sln
@@ -66,6 +66,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SchemaManager.Core.UnitTest
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.SqlServer.Tests.Integration", "test\Microsoft.Health.SqlServer.Tests.Integration\Microsoft.Health.SqlServer.Tests.Integration.csproj", "{5429C3A8-3429-43EA-BA05-E347CB4D1F61}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Health.Checkpoints", "src\Microsoft.Health.Checkpoints\Microsoft.Health.Checkpoints.csproj", "{6D06503C-E79B-4F05-AFE1-C25386220895}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -172,6 +174,10 @@ Global
 		{5429C3A8-3429-43EA-BA05-E347CB4D1F61}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{5429C3A8-3429-43EA-BA05-E347CB4D1F61}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{5429C3A8-3429-43EA-BA05-E347CB4D1F61}.Release|Any CPU.Build.0 = Release|Any CPU
+		{6D06503C-E79B-4F05-AFE1-C25386220895}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{6D06503C-E79B-4F05-AFE1-C25386220895}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{6D06503C-E79B-4F05-AFE1-C25386220895}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{6D06503C-E79B-4F05-AFE1-C25386220895}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -202,9 +208,10 @@ Global
 		{72816760-BDE2-4CAF-AF41-F7BEE8E26158} = {B70945F4-01A6-4351-955B-C4A2943B5E3B}
 		{0BEB2EE3-DBC3-4DC4-BB01-141DBFFED8D2} = {CCD9FF99-E177-446E-B9E5-9F570FD96A34}
 		{5429C3A8-3429-43EA-BA05-E347CB4D1F61} = {CCD9FF99-E177-446E-B9E5-9F570FD96A34}
+		{6D06503C-E79B-4F05-AFE1-C25386220895} = {8AD2A324-DAB5-4380-94A5-31F7D817C384}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {562bbd07-1817-4af5-ab66-8837cf849248}
 		RESX_SortFileContentOnSave = True
+		SolutionGuid = {562bbd07-1817-4af5-ab66-8837cf849248}
 	EndGlobalSection
 EndGlobal
--- a/src/Microsoft.Health.Checkpoints/Checkpoint.cs
+++ b/src/Microsoft.Health.Checkpoints/Checkpoint.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Health.Checkpoints
 
         public string Identifier { get; set; }
 
-        public Azure.ETag ETag { get; set; }
+        public string ETag { get; set; }
 
         public DateTimeOffset LastProcessedDateTime { get; set; }
 
--- a/src/Microsoft.Health.Checkpoints/Checkpoint.cs
+++ b/src/Microsoft.Health.Checkpoints/Checkpoint.cs
@@ -9,10 +9,14 @@ namespace Microsoft.Health.Checkpoints
 {
     public class Checkpoint : ICheckpoint
     {
+        public string Partition { get; set; }
+
         public string Identifier { get; set; }
 
-        public DateTimeOffset LastProcessedDt { get; set; }
+        public Azure.ETag ETag { get; set; }
 
-        public string LastProcessedId { get; set; }
+        public DateTimeOffset LastProcessedDateTime { get; set; }
+
+        public string LastProcessedIdentifier { get; set; }
     }
 }
--- a/src/Microsoft.Health.Checkpoints/Checkpoint.cs
+++ b/src/Microsoft.Health.Checkpoints/Checkpoint.cs
@@ -1,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Checkpoints
+{
+    public class Checkpoint : ICheckpoint
+    {
+        public string Identifier { get; set; }
+
+        public DateTimeOffset LastProcessedDt { get; set; }
+
+        public string LastProcessedId { get; set; }
+    }
+}
--- a/src/Microsoft.Health.Checkpoints/ICheckpoint.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpoint.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Health.Checkpoints
 
         string Identifier { get; set; }
 
-        Azure.ETag ETag { get; set; }
+        string ETag { get; set; }
 
         DateTimeOffset LastProcessedDateTime { get; set; }
 
--- a/src/Microsoft.Health.Checkpoints/ICheckpoint.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpoint.cs
@@ -9,10 +9,14 @@ namespace Microsoft.Health.Checkpoints
 {
     public interface ICheckpoint
     {
+        string Partition { get; set; }
+
         string Identifier { get; set; }
 
-        DateTimeOffset LastProcessedDt { get; set; }
+        Azure.ETag ETag { get; set; }
 
-        string LastProcessedId { get; set; }
+        DateTimeOffset LastProcessedDateTime { get; set; }
+
+        string LastProcessedIdentifier { get; set; }
     }
 }
--- a/src/Microsoft.Health.Checkpoints/ICheckpoint.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpoint.cs
@@ -1,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Checkpoints
+{
+    public interface ICheckpoint
+    {
+        string Identifier { get; set; }
+
+        DateTimeOffset LastProcessedDt { get; set; }
+
+        string LastProcessedId { get; set; }
+    }
+}
--- a/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
@@ -11,6 +11,6 @@ namespace Microsoft.Health.Checkpoints
     {
         Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier);
 
-        Task SetCheckpointAsync(ICheckpoint checkpoint);
+        Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint);
     }
 }
--- a/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Health.Checkpoints
 {
     public interface ICheckpointClient
     {
-        Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier);
+        Task<ICheckpoint> GetCheckpointAsync(string partition, string checkpointIdentifier);
 
         Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint);
     }
--- a/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
@@ -1,0 +1,16 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Checkpoints
+{
+    public interface ICheckpointClient
+    {
+        Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier);
+
+        Task SetCheckpointAsync(ICheckpoint checkpoint);
+    }
+}
--- a/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
+++ b/src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
@@ -3,14 +3,15 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Health.Checkpoints
 {
     public interface ICheckpointClient
     {
-        Task<ICheckpoint> GetCheckpointAsync(string partition, string checkpointIdentifier);
+        Task<ICheckpoint> GetCheckpointAsync(string partition, string checkpointIdentifier, CancellationToken token = default);
 
-        Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint);
+        Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint, CancellationToken token = default);
     }
 }
--- a/src/Microsoft.Health.Checkpoints/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.Health.Checkpoints/Properties/AssemblyInfo.cs
@@ -1,0 +1,10 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Resources;
+
+[assembly: CLSCompliant(false)]
+[assembly: NeutralResourcesLanguage("en-us")]
--- a/src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
+++ b/src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
@@ -7,76 +7,120 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Azure;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using EnsureThat;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Health.Checkpoints.Storage
 {
     public class BaseStorageCheckpointClient : ICheckpointClient
     {
         private BlobContainerClient _storageClient;
+        private ILogger<BaseStorageCheckpointClient> _logger;
+        private const string _lastProcessedDateTime = "LastProcessedDateTime";
+        private const string _lastProcessedIdentifier = "LastProcessedIdentifier";
 
-        public BaseStorageCheckpointClient(BlobContainerClient containerClient)
+        public BaseStorageCheckpointClient(BlobContainerClient containerClient, ILogger<BaseStorageCheckpointClient> logger)
         {
             EnsureArg.IsNotNull(containerClient);
 
             _storageClient = containerClient;
+            _logger = logger;
         }
 
-        public virtual Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier)
+        public virtual async Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier)
         {
-            Task<ICheckpoint> GetCheckpointAsync()
+            EnsureArg.IsNotNullOrEmpty(checkpointIdentifier, nameof(checkpointIdentifier));
+
+            ICheckpoint checkpoint = new Checkpoint();
+
+            try
             {
-                ICheckpoint checkpoint = new Checkpoint();
+                var resultSegment = _storageClient.GetBlobsAsync(
+                                        traits: BlobTraits.Metadata,
+                                        states: BlobStates.All,
+                                        prefix: checkpointIdentifier,
+                                        cancellationToken: CancellationToken.None)
+                                        .AsPages();
 
-                foreach (BlobItem blob in _storageClient.GetBlobs(traits: BlobTraits.Metadata, states: BlobStates.All, prefix: checkpointIdentifier, cancellationToken: CancellationToken.None))
+                await foreach (Page<BlobItem> blobPage in resultSegment)
                 {
-                    DateTimeOffset lastEventTimestamp = DateTime.MinValue;
-
-                    if (blob.Metadata.TryGetValue("LastProcessedDt", out var dt))
+                    if (blobPage.Values.Count == 0)
                     {
-                        DateTimeOffset.TryParse(dt, null, DateTimeStyles.AssumeUniversal, out lastEventTimestamp);
-                        checkpoint.LastProcessedDt = lastEventTimestamp;
+                        _logger.LogWarning($"No blob found for identifier {checkpointIdentifier}");
                     }
 
-                    if (blob.Metadata.TryGetValue("LastProcessedId", out var id))
+                    if (blobPage.Values.Count > 1)
                     {
-                        checkpoint.LastProcessedId = id;
+                        _logger.LogWarning($"Multiple blobs found for identifier {checkpointIdentifier}");
                     }
 
-                    checkpoint.Identifier = checkpointIdentifier;
+                    foreach (BlobItem blobItem in blobPage.Values)
+                    {
+                        DateTimeOffset lastEventTimestamp = DateTime.MinValue;
+
+                        if (blobItem.Metadata.TryGetValue(_lastProcessedDateTime, out var dt))
+                        {
+                            DateTimeOffset.TryParse(dt, null, DateTimeStyles.AssumeUniversal, out lastEventTimestamp);
+                            checkpoint.LastProcessedDateTime = lastEventTimestamp;
+                        }
+
+                        if (blobItem.Metadata.TryGetValue(_lastProcessedIdentifier, out var id))
+                        {
+                            checkpoint.LastProcessedIdentifier = id;
+                        }
+
+                        if (checkpoint.LastProcessedDateTime == DateTime.MinValue || checkpoint.LastProcessedIdentifier == null)
+                        {
+                            _logger.LogWarning($"No valid checkpoint found for {checkpointIdentifier}. Using default checkpoint of {checkpoint.LastProcessedDateTime}");
+                        }
+
+                        checkpoint.ETag = (ETag)blobItem.Properties.ETag;
+                        checkpoint.Identifier = checkpointIdentifier;
+                    }
                 }
-
-                return Task.FromResult(checkpoint);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, ex.Message);
+                throw;
             }
 
-            return GetCheckpointAsync();
+            return checkpoint;
         }
 
-        public virtual async Task SetCheckpointAsync(ICheckpoint checkpoint)
+        public virtual async Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint)
         {
             EnsureArg.IsNotNull(checkpoint);
             EnsureArg.IsNotNullOrWhiteSpace(checkpoint.Identifier);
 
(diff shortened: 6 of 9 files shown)


src/Microsoft.Health.Checkpoints/ICheckpoint.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.Health.Checkpoints
+{
+    public interface ICheckpoint
+    {
+        string Identifier { get; set; }
+
+        DateTimeOffset LastProcessedDt { get; set; }
+
+        string LastProcessedId { get; set; }
+    }
```
dustinburson, line 17: Suggestion: Thoughts on adding a partition concept here and using it in the path to the underlying blob?  Right now we will only have 1 but in the future we will likely have multiple agents each with their own affinity.

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
```diff
@@ -0,0 +1,90 @@
...
+            _storageClient = containerClient;
+        }
+
+        public virtual Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier)
+        {
+            Task<ICheckpoint> GetCheckpointAsync()
+            {
+                ICheckpoint checkpoint = new Checkpoint();
+
+                foreach (BlobItem blob in _storageClient.GetBlobs(traits: BlobTraits.Metadata, states: BlobStates.All, prefix: checkpointIdentifier, cancellationToken: CancellationToken.None))
```
dustinburson, line 36: This is a foreach because there isn't just a GetBlob API call? If so the expectation is we should only have one blob found correct? In that case we should either have an error condition if more than one is found or at least a warning that is logged.

Also, is there a reason we aren't using the GetBlobsAsync?  We should use the async call when possible so the thread pool thread isn't blocked during I/O and returned to the pool for other work.

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    DateTimeOffset lastEventTimestamp = DateTime.MinValue;
+
+                    if (blob.Metadata.TryGetValue("LastProcessedDt", out var dt))
```
dustinburson, line 40: Should this be an error condition? If the checkpoint is empty or partially empty is it still valid?  Would at a minimum suggest injecting an ILogger and log a warning.

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            var metadata = new Dictionary<string, string>()
+            {
+                { "LastProcessedDt",  lastProcessedDt },
+                { "LastProcessedId",  lastProcessedId },
+            };
+
+            try
+            {
+                await blobClient.SetMetadataAsync(metadata);
```
dustinburson, line 79: We should probably use BlobRequestConditions and include the ETag in the ICheckpoint for optimistic concurrency via IfMatch.  Right now there *should* only be one client updating but this should help protect against concurrency errors if something ends up being misconfigured.

See https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.models.blobrequestconditions?view=azure-dotnet

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+            catch (RequestFailedException ex) when ((ex.ErrorCode == BlobErrorCode.BlobNotFound) || (ex.ErrorCode == BlobErrorCode.ContainerNotFound))
```
dustinburson, line 81: Interested in knowing the thinking behind not using ExistsAsync https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobcontainerclient.existsasync.  Is the thought to optimized for the normal scenario where the checkpoint exists and rather than always doing two blob client calls (Exists & SetMetadata)?

dustinburson (changes requested):

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
dustinburson: This could be handled in how the identifier is generated but it seems cleaner to include it natively.

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System;
+
+namespace Microsoft.Health.Checkpoints
+{
+    public interface ICheckpoint
+    {
+        string Identifier { get; set; }
+
+        DateTimeOffset LastProcessedDt { get; set; }
```
brendankowitz, line 14: nit: avoid the acronym in the property name

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
```diff
@@ -0,0 +1,90 @@
...
+        private BlobContainerClient _storageClient;
+
+        public BaseStorageCheckpointClient(BlobContainerClient containerClient)
+        {
+            EnsureArg.IsNotNull(containerClient);
+
+            _storageClient = containerClient;
+        }
+
+        public virtual Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier)
```
brendankowitz, line 30: Nit: Can this be null? If yes, perhaps add a default value `checkpointIdentifier = null`. If not add `EnsureArg.IsNotNullOrEmpty()`

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
brendankowitz: nit: `LastProcessedDt` is used several times and could be a const

brandonpollett (commented):
What's the overlap between what's here and what is in the [Microsoft.Health.Blob](https://github.com/microsoft/healthcare-shared-components/tree/master/src/Microsoft.Health.Blob) project? Is the idea to in the future have different providers of the checkpoint or is it always going to be blob?

Microsoft.Health.sln
```diff
@@ -66,6 +66,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SchemaManager.Core.UnitTest
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.SqlServer.Tests.Integration", "test\Microsoft.Health.SqlServer.Tests.Integration\Microsoft.Health.SqlServer.Tests.Integration.csproj", "{5429C3A8-3429-43EA-BA05-E347CB4D1F61}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Health.Checkpoints", "src\Microsoft.Health.Checkpoints\Microsoft.Health.Checkpoints.csproj", "{6D06503C-E79B-4F05-AFE1-C25386220895}"
```
ms-teli, line 69: Question: Is this checkpoint designed as a generic solution based off the blob storage in this version? For example, we can support Event and IoT Connector?

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
wi-y: I had partition originally but realized it was kind of redundant since the the identifier can include the partition. But will add partition back so that the two concepts of partition and id are obvious to the user.

Microsoft.Health.sln
wi-y: Yes the goal for this PR is two define the generic classes to be able to support checkpointing in the Events and IoT connector projects. I had taken the checkpointing that exists for IoT connector and basically added it here with the intent of supporting both the Events use cases and the IoT Connector use cases which are pretty similar. It should support other uses cases as well. I will eventually deprecate the IoT Connector checkpointing classes and replace them with this.

wi-y:
> What's the overlap between what's here and what is in the [Microsoft.Health.Blob](https://github.com/microsoft/healthcare-shared-components/tree/master/src/Microsoft.Health.Blob) project? Is the idea to in the future have different providers of the checkpoint or is it always going to be blob?

@brandonpollett 
Yeah the goal is to have some sort of checkpointing mechanism but this mechanism is not necessarily tied to blob. The checkpoints could be stored in a SQL db, or Cosmos in the future, but for now blob is what we are going to write a proof of concept for, so we only have an example class for blob at the moment. Since there may be future cases that are not blob I decided to put the checkpointing classes/interfaces in a separate project from Microsoft.Health.Blob.

I did verify that we can leverage Microsoft.Health.Blob to use services.AddBlobDataStore() and use BlobServiceClient to get a BlobContainerClient which can be injected into the BaseStorageCheckpointClient in this lib.

brandonpollett:
> > What's the overlap between what's here and what is in the [Microsoft.Health.Blob](https://github.com/microsoft/healthcare-shared-components/tree/master/src/Microsoft.Health.Blob) project? Is the idea to in the future have different providers of the checkpoint or is it always going to be blob?
> 
> @brandonpollett
> Yeah the goal is to have some sort of checkpointing mechanism but this mechanism is not necessarily tied to blob. The checkpoints could be stored in a SQL db, or Cosmos in the future, but for now blob is what we are going to write a proof of concept for, so we only have an example class for blob at the moment. Since there may be future cases that are not blob I decided to put the checkpointing classes/interfaces in a separate project from Microsoft.Health.Blob.
> 
> I did verify that we can leverage Microsoft.Health.Blob to use services.AddBlobDataStore() and use BlobServiceClient to get a BlobContainerClient which can be injected into the BaseStorageCheckpointClient in this lib.

Cool. In the future we'll probably want to move the blob implementation out of the checkpoint library so that if somebody wants to use a SQL or Cosmos version, they don't have to take a dependency on the blob libraries.

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
wi-y: There does not happen to be a GetBlob API call. Updated with logging and changed to GetBlobsAsync

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
wi-y: Updated to log a warning and now use a const

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
wi-y: Makes sense. Added ETag support

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
wi-y: It seemed better than always making two blob client calls, but the real reason is that this code was shamelessly "borrowed" from the event hubs checkpointing code

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
wi-y: updated

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
wi-y: yep, thanks. updated to use EnsureArg.IsNotNullOrEmpty()

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
wi-y: I am still on the fence about including a separate property for partition vs. just have an identifier with includes the partition + id. 

I would prefer to have one argument in ICheckpointClient here:
Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier);

If we add partition, then I think we'd have to add an additional argument to the ICheckpointClient interface:
Task<ICheckpoint> GetCheckpointAsync(string checkpointIdentifier, string partitionId);

But then we'd just end up concat-ing them later when we do the blob lookup. Is updating the interface to have two argument preferred?

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Threading.Tasks;
+using Azure;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using EnsureThat;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Health.Checkpoints.Storage
+{
+    public class BaseStorageCheckpointClient : ICheckpointClient
```
dustinburson, line 21: Typically we only use Base if this is going to be abstract and not used as is.  Would suggest renaming StorageCheckpointClient, it will be the default implementation.  If others build on top of they will be named differently.

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
dustinburson: I guess the real question is how will this be used up stream.  We could remove partition as a concept in the ICheckpoint itself.  Instead we could have another abstraction on top of the ICheckpointClient, i.e. PartitionedStorageCheckpointClient and potentially an additional interface, IParitionedCheckpointClient that introduces a required PartitionId property.  This class would have the responsibility of appending the partition id to the base identifier.  Thoughts?  If that approach sounds good we don't need to add it yet but is probably something we will add once we build on top of this for the IoT Connector version.

My main concern is I don't want partition management to be arbitrary string concats from the code calling this client.  When we do handle partitions I would like it to be well defined.

src/Microsoft.Health.Checkpoints/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Resources;
+
+[assembly: CLSCompliant(false)]
```
dustinburson, line 9: Out of curiosity what prevents this from being CLSCompliant(true)?

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
dustinburson: That is fine and makes sense.  One suggestion, if we are already using Polly elsewhere in the Shared repo it might be worth leveraging that library here for a slightly cleaner implementation.  Look at line 70 in https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs for an example.  Here we handle an optimistic concurrency failure with a refresh of the data from the server and a fresh merge with the internal data.

src/Microsoft.Health.Checkpoints/Checkpoint.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Health.Checkpoints
+{
+    public class Checkpoint : ICheckpoint
+    {
+        public string Partition { get; set; }
+
+        public string Identifier { get; set; }
+
+        public Azure.ETag ETag { get; set; }
```
dustinburson, line 16: I don't know if we need to have this as the Azure.ETag object.  Can we simplify the interface and represent this as a string and interpolate between the two internally?

src/Microsoft.Health.Checkpoints/ICheckpoint.cs
wi-y: I went ahead and added partition as part of the interface here so that we don't need to do another PartitionedStorageCheckpointClient in the future.

src/Microsoft.Health.Checkpoints/Storage/BaseStorageCheckpointClient.cs
wi-y: updated name

src/Microsoft.Health.Checkpoints/Properties/AssemblyInfo.cs
wi-y: I am not sure if there is anything specific that makes it not compliant other than us not verifying that it is compliant. The rest of the projects in this solution did this, so just wanted to be consistent.

src/Microsoft.Health.Checkpoints/Checkpoint.cs
wi-y: yep, updated to string.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            EnsureArg.IsNotNullOrEmpty(checkpointIdentifier, nameof(checkpointIdentifier));
+
+            ICheckpoint checkpoint = new Checkpoint();
+
+            try
+            {
+                var resultSegment = _storageClient.GetBlobsAsync(
+                                        traits: BlobTraits.Metadata,
+                                        states: BlobStates.All,
+                                        prefix: $"{partition}/{checkpointIdentifier}",
```
dustinburson, line 48: Suggestion:  This is repeated in the SetCheckpointAsync,  would be better to have at least a private static method use in both places to ensure the prefix is built the same no matter what.  If you are concerned with method call stack overhead you can use MethoImpl aggressive inlining so the compiler inline the contents of the method (functionally replicating what you have here).

dustinburson (approved):

src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Checkpoints
+{
+    public interface ICheckpointClient
+    {
+        Task<ICheckpoint> GetCheckpointAsync(string partition, string checkpointIdentifier);
```
brandonpollett, line 12: nit: cancellation token support on async methods (and interfaces) something like `CancellationToken token = default`.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            _storageClient = containerClient;
+            _logger = logger;
+        }
+
+        public virtual async Task<ICheckpoint> GetCheckpointAsync(string partition, string checkpointIdentifier)
+        {
+            EnsureArg.IsNotNullOrEmpty(partition, nameof(partition));
+            EnsureArg.IsNotNullOrEmpty(checkpointIdentifier, nameof(checkpointIdentifier));
+
+            ICheckpoint checkpoint = new Checkpoint();
```
brandonpollett, line 41: nit: use var when the type is explicit

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            try
+            {
+                var resultSegment = _storageClient.GetBlobsAsync(
```
brandonpollett, line 45: nit: use explicit type when the type is not clear from method.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                        }
+
+                        checkpoint.ETag = blobItem.Properties.ETag.ToString();
+                        checkpoint.Identifier = checkpointIdentifier;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, ex.Message);
```
brandonpollett, line 91: Is this needed? Generally wouldn't this would be captured up the stack also and logged?

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                var resultSegment = _storageClient.GetBlobsAsync(
+                                        traits: BlobTraits.Metadata,
+                                        states: BlobStates.All,
+                                        prefix: $"{partition}/{checkpointIdentifier}",
+                                        cancellationToken: CancellationToken.None)
+                                        .AsPages();
+
+                await foreach (Page<BlobItem> blobPage in resultSegment)
+                {
+                    if (blobPage.Values.Count == 0)
```
brandonpollett, line 54: Does this mean that the code that uses this needs to know what to do with an empty checkpoint? Would it make more sense to return null?

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                    {
+                        _logger.LogWarning($"No blob found for identifier {checkpointIdentifier}");
+                    }
+
+                    if (blobPage.Values.Count > 1)
```
brandonpollett, line 59: Does this mean that the last one processed wins? Is that the behavior that we want? Also, looking at the implementation, it seems like we could get a combination of many different checkpoints if the `TryParse` commands below fail.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                    {
+                        _logger.LogWarning($"Multiple blobs found for identifier {checkpointIdentifier}");
+                    }
+
+                    foreach (BlobItem blobItem in blobPage.Values)
+                    {
+                        DateTimeOffset lastEventTimestamp = DateTime.MinValue;
+
+                        if (blobItem.Metadata.TryGetValue(_lastProcessedDateTime, out var dt))
```
brandonpollett, line 68: What type is `dt`? explicit type is probably more readable.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                            DateTimeOffset.TryParse(dt, null, DateTimeStyles.AssumeUniversal, out lastEventTimestamp);
+                            checkpoint.LastProcessedDateTime = lastEventTimestamp;
+                        }
+
+                        if (blobItem.Metadata.TryGetValue(_lastProcessedIdentifier, out var id))
+                        {
+                            checkpoint.LastProcessedIdentifier = id;
+                        }
+
+                        if (checkpoint.LastProcessedDateTime == DateTime.MinValue || checkpoint.LastProcessedIdentifier == null)
```
brandonpollett, line 79: Probably would make more sense to use `checkpoint.LastProcessedDateTime == default`. Also, we would generally use something like `string.IsNullOrEmpty` or `string.IsNullOrWhitespace` if they're appropriate here.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            var metadata = new Dictionary<string, string>()
+            {
+                { _lastProcessedDateTime,  lastProcessedDateTime },
+                { _lastProcessedIdentifier,  lastProcessedIdentifier },
+            };
+
+            try
+            {
+                var blobRequestOptions = new BlobRequestConditions();
```
brandonpollett, line 118: nit: use object initialization.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            return checkpoint;
+        }
+
+        public virtual async Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint)
+        {
+            EnsureArg.IsNotNull(checkpoint);
+            EnsureArg.IsNotNullOrWhiteSpace(checkpoint.Partition);
+            EnsureArg.IsNotNullOrWhiteSpace(checkpoint.Identifier);
+
+            var lastProcessedDateTime = EnsureArg.IsNotNullOrWhiteSpace(checkpoint.LastProcessedDateTime.DateTime.ToString("MM/dd/yyyy hh:mm:ss.fff tt"), nameof(checkpoint.LastProcessedDateTime));
```
brandonpollett, line 104: `LastProcessedTime` is a `DateTimeOffset` (so struct) and it can't be null. You can probably safely remove this and below just use `checkpoint.LastProcessedDateTime.ToString("MM/dd/yyyy hh:mm:ss.fff tt")`.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                { _lastProcessedDateTime,  lastProcessedDateTime },
+                { _lastProcessedIdentifier,  lastProcessedIdentifier },
+            };
+
+            try
+            {
+                var blobRequestOptions = new BlobRequestConditions();
+                blobRequestOptions.IfMatch = new ETag(checkpoint.ETag);
+                BlobInfo result = await blobClient.SetMetadataAsync(metadata, blobRequestOptions);
```
brandonpollett, line 120: Cancellation support?

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                checkpoint.ETag = result.ETag.ToString();
+            }
+            catch (RequestFailedException ex) when (ex.Status == (int)HttpStatusCode.PreconditionFailed)
+            {
+                _logger.LogWarning("The checkpoint's ETag does not match ETag provided.");
+            }
+            catch (RequestFailedException ex) when ((ex.ErrorCode == BlobErrorCode.BlobNotFound) || (ex.ErrorCode == BlobErrorCode.ContainerNotFound))
```
brandonpollett, line 127: I believe the parentheses around each side of the or are redundant.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                using (var blobContent = new MemoryStream(Array.Empty<byte>()))
+                {
+                    BlobContentInfo result = await blobClient.UploadAsync(blobContent, metadata: metadata).ConfigureAwait(false);
+                    checkpoint.ETag = result.ETag.ToString();
+                }
+            }
+            catch (Exception ex)
```
brandonpollett, line 135: We would generally not swallow the generic exception.

src/Microsoft.Health.Checkpoints/Properties/AssemblyInfo.cs
wsugarman: You can make this compliant if you wish, and you can read about the requirements [here](https://docs.microsoft.com/en-us/dotnet/standard/language-independence).

Ultimately, you'll probably just be limited by the types you can use (like no unsigned types). When I added the assembly attributes, I think each of the assemblies at the time had some conformance problem (at least as I remember it).

src/Microsoft.Health.Checkpoints/ICheckpointClient.cs
wi-y: updated

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: updated

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: No, not really. It can be captured elsewhere. Removed

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: null would make more sense, thanks. updated

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: In practice there will only be one blob per partition with checkpoint information.

The blob api makes use specify a "prefix" when searching for a blob. In our case the prefix is just the full blob path so we would only expect 0 or 1 result. In the result, there will be either a lastProcessedDateTime or lastProcessedIdentifier (or both). 

If there is more than 1 result, then that is not expected and we throw an exception.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: updated

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -33,78 +33,74 @@ public StorageCheckpointClient(BlobContainerClient containerClient, ILogger<Stor
...
-            ICheckpoint checkpoint = new Checkpoint();
+            var checkpoint = new Checkpoint();
+            var blobName = GetBlobName(partition, checkpointIdentifier);
 
-            try
+            IAsyncEnumerable<Page<BlobItem>> resultSegment = _storageClient.GetBlobsAsync(
+                                    traits: BlobTraits.Metadata,
+                                    states: BlobStates.All,
+                                    prefix: blobName,
+                                    cancellationToken: CancellationToken.None)
```
brandonpollett, line 48: We should pass in `token` here.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -33,78 +33,74 @@ public StorageCheckpointClient(BlobContainerClient containerClient, ILogger<Stor
...
 
-        public virtual async Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint)
+        public virtual async Task<ICheckpoint> SetCheckpointAsync(ICheckpoint checkpoint, CancellationToken token = default)
         {
             EnsureArg.IsNotNull(checkpoint);
             EnsureArg.IsNotNullOrWhiteSpace(checkpoint.Partition);
             EnsureArg.IsNotNullOrWhiteSpace(checkpoint.Identifier);
 
-            var lastProcessedDateTime = EnsureArg.IsNotNullOrWhiteSpace(checkpoint.LastProcessedDateTime.DateTime.ToString("MM/dd/yyyy hh:mm:ss.fff tt"), nameof(checkpoint.LastProcessedDateTime));
+            var lastProcessedDateTime = checkpoint.LastProcessedDateTime.ToString("MM/dd/yyyy hh:mm:ss.fff tt");
```
brandonpollett, line 100: Any particular reason to not use one of the standard ones? https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings

brandonpollett: Also, in this method, won't we be stripping away any offset that is set by the consumer and then possibly skewing the time since on the way out we assume universal? Seems like something like https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#Roundtrip maybe more appropriate.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
```diff
@@ -33,78 +33,74 @@ public StorageCheckpointClient(BlobContainerClient containerClient, ILogger<Stor
...
-                                        traits: BlobTraits.Metadata,
-                                        states: BlobStates.All,
-                                        prefix: $"{partition}/{checkpointIdentifier}",
-                                        cancellationToken: CancellationToken.None)
-                                        .AsPages();
-
-                await foreach (Page<BlobItem> blobPage in resultSegment)
+                if (blobPage.Values.Count == 0)
+                {
+                    _logger.LogInformation($"No blob found for blob name {blobName}");
```
brandonpollett, line 55: For structured logging to work, we should not use string interpolation. (Also elsewhere in this PR). The line should be something like:

```csharp
_logger.LogInformation("No blob found for blob name {blobName}", blobName);
```

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: That is a great point. Yes, roundtrip is a much better approach. The previous approach would have skewed the time. Updated to use roundtrip.

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: yep, missed that one. updated

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClient.cs
wi-y: updated

src/Microsoft.Health.Checkpoints/Storage/StorageCheckpointClientException.cs
```diff
@@ -0,0 +1,26 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Checkpoints.Storage
+{
+    public class StorageCheckpointClientException : Exception
```
brandonpollett, line 10: We might want to extend this from `MicrosoftHealthException`

brandonpollett (approved):

PR #260, opened 2021-07-01
Add ProcessTerminator

YazanMSFT:
Add config for terminating the service when a specific schema version is reached
Add tests

## Description
These changes are needed for some schema upgrade scenarios.

## Related issues
Addresses [issue #].

## Testing
Unit tests were added, also tested using the test web project.

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
@@ -11,6 +11,7 @@ using MediatR;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Health.Core;
+using Microsoft.Health.Core.Features.Control;
 using Microsoft.Health.SqlServer.Configs;
 using Microsoft.Health.SqlServer.Features.Schema.Extensions;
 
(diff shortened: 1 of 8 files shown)


rbans96 (approved):

src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -26,22 +27,30 @@ public class SchemaJobWorker
...
         {
             EnsureArg.IsNotNull(services, nameof(services));
             EnsureArg.IsNotNull(sqlServerDataStoreConfiguration, nameof(sqlServerDataStoreConfiguration));
             EnsureArg.IsNotNull(mediator, nameof(mediator));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
+            if (sqlServerDataStoreConfiguration.TerminateWhenSchemaVersionUpdatedTo.HasValue)
+            {
+                EnsureArg.IsNotNull(processTerminator, nameof(processTerminator));
+            }
```
Ivanidzo4ka, line 48: It doesn't work this way unfortunatelly.

You will get 

Unable to resolve service for type 'Microsoft.Health.Core.Features.Control.IProcessTerminator' while attempting to activate 'Microsoft.Health.SqlServer.Features.Schema.SchemaJobWorker'.


if you try to run it without regestring process terminator


As option, we can have NoopProcessTerminator class which does nothing and register it based on config.

PR #261, opened 2021-07-02
Update Blob Client APIs

wsugarman:
## Description
- Refactor the APIs for registering `BlobServiceClient` services in the service container
  - Provide separate APIs for registering a client and registering an `IHostedService` for initializing the blob container
- Refactor configuration use to leverage `IOptions` for better DI integration
- `Obsolete` the previous API

## Related issues
[AB#83291](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83291)

## Testing
Run unit tests

Feature


src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -29,43 +33,132 @@ public static IServiceCollection AddBlobDataStore(this IServiceCollection servic
...
+                        .GetService<IConfiguration>()
+                        .GetSection(BlobDataStoreConfiguration.SectionName)
+                        .Bind(config);
 
-                    if (string.IsNullOrEmpty(config.ConnectionString) && config.AuthenticationType == BlobDataStoreAuthenticationType.ConnectionString)
-                    {
-                        config.ConnectionString = BlobLocalEmulator.ConnectionString;
-                    }
+                    configureAction?.Invoke(config);
+                    DefaultBlobDataStoreConfiguration.Instance.Configure(config);
```
pengchen0692, line 46: seems we could call `services.ConfigureOptions<DefaultBlobDataStoreConfiguration>()` outside this singleton call, then options system can automatically configure it.

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -29,43 +33,132 @@ public static IServiceCollection AddBlobDataStore(this IServiceCollection servic
...
+        /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null"/>.</exception>
+        public static IServiceCollection AddBlobServiceClient(this IServiceCollection services)
+        {
+            EnsureArg.IsNotNull(services, nameof(services));
 
-            services.Add<BlobClientReadWriteTestProvider>()
-                .Singleton()
-                .AsService<IBlobClientTestProvider>();
+            services.AddOptions();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<BlobDataStoreConfiguration>>(new DefaultBlobDataStoreConfiguration()));
```
pengchen0692, line 67: could we simplify this by using `ConfigureOptions`?
reference: https://andrewlock.net/simplifying-dependency-injection-for-iconfigureoptions-with-the-configureoptions-helper/

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -29,43 +33,132 @@ public static IServiceCollection AddBlobDataStore(this IServiceCollection servic
...
+        /// <returns>The <paramref name="services"/> for additional method invocations.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null"/>.</exception>
+        public static IServiceCollection AddBlobServiceClient(this IServiceCollection services)
+        {
+            EnsureArg.IsNotNull(services, nameof(services));
 
-            services.Add<BlobClientReadWriteTestProvider>()
-                .Singleton()
-                .AsService<IBlobClientTestProvider>();
+            services.AddOptions();
```
pengchen0692, line 66: should we have `AddOptions` here or should it be put in some more common place.

src/Microsoft.Health.Blob/Registration/DefaultBlobDataStoreConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Blob.Configs;
+using Microsoft.Health.Blob.Features.Storage;
+
+namespace Microsoft.Health.Blob.Registration
+{
+    internal class DefaultBlobDataStoreConfiguration : IConfigureOptions<BlobDataStoreConfiguration>
```
pengchen0692, line 12: Futuremore, I'm thinking we probably could replace  `Action<T> configure` when configure with `IConfigureOptions` in the future.

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -29,43 +33,132 @@ public static IServiceCollection AddBlobDataStore(this IServiceCollection servic
...
+        /// is already initialized by the time service can be resolved.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to be updated.</param>
+        /// <returns>The <paramref name="services"/> for additional method invocations.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null"/>.</exception>
+        public static IServiceCollection AddInitializedBlobServiceClient(this IServiceCollection services)
+        {
+            EnsureArg.IsNotNull(services, nameof(services));
+
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<BlobDataStoreConfiguration>>(new DefaultBlobDataStoreConfiguration()));
```
pengchen0692, line 102: could we use [COnfigureOptions](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.configureoptions-1?view=dotnet-plat-ext-5.0) to simplify this?

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
wsugarman: We should keep it here. People can repeatedly call `AddOptions` without issue

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
wsugarman: I think that comes down to whether want to automatically read the configuration by default and bind the options to it. My inclination is "No," and for clients to perform the binding themselves based on their own configs. Then to that end we don't want all of the `IConfigureOptions` classes to register automatically.

pengchen0692: I'm thinking if we registered multiple implementation of `IConfigureOptions<BlobDataStoreConfiguration>`, which one will be used.

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -29,43 +30,55 @@ public static IServiceCollection AddBlobDataStore(this IServiceCollection servic
...
-                .Singleton()
-                .AsSelf();
+            services.AddOptions();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<BlobDataStoreConfiguration>>(new DefaultBlobDataStoreConfiguration()));
+            services.TryAddSingleton(p => BlobClientFactory.Create(p.GetRequiredService<IOptions<BlobDataStoreConfiguration>>().Value));
 
-            services.Add<BlobClientReadWriteTestProvider>()
-                .Singleton()
-                .AsService<IBlobClientTestProvider>();
+            return new BlobClientBuilder(services);
```
smithago, line 62: Is this enough? The containers are not initialized right?

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -29,43 +30,55 @@ public static IServiceCollection AddBlobDataStore(this IServiceCollection servic
...
+        {
+            EnsureArg.IsNotNull(configure, nameof(configure));
 
-            services.TryAddSingleton<RecyclableMemoryStreamManager>();
+            BlobClientBuilder builder = services.AddBlobServiceClient();
+            services.Configure(configure);
 
-            return services;
+            return builder;
         }
```
smithago, line 82: There are so many public methods. It will be confusing to know which one to use. 

All we need is 2 right. Both need to initialize everything except in different workflow

src/Microsoft.Health.Blob/Registration/DefaultBlobDataStoreConfiguration.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Blob.Configs;
+using Microsoft.Health.Blob.Features.Storage;
+
+namespace Microsoft.Health.Blob.Registration
+{
+    internal class DefaultBlobDataStoreConfiguration : IPostConfigureOptions<BlobDataStoreConfiguration>
+    {
+        public void PostConfigure(string name, BlobDataStoreConfiguration options)
```
smithago, line 14: asp.net already manages the instances for us, can we use that instead?

smithago: Will this change independently of asp.net maintained one when we use IOPtionSnapshot?

src/Microsoft.Health.Blob/Features/Storage/BlobClientInitializer.cs
```diff
@@ -34,21 +33,7 @@ public BlobServiceClient CreateBlobClient(BlobDataStoreConfiguration configurati
...
-            blobClientOptions.Retry.NetworkTimeout = TimeSpan.FromMinutes(configuration.RequestOptions.ServerTimeoutInMinutes);
-
-            if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
-            {
-                var defaultCredentials = new DefaultAzureCredential();
-                return new BlobServiceClient(new Uri(configuration.ConnectionString), defaultCredentials, blobClientOptions);
-            }
-
-            return new BlobServiceClient(configuration.ConnectionString, blobClientOptions);
+            return BlobClientFactory.Create(configuration);
```
smithago, line 36: Not using DI here

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -5,21 +5,22 @@
...
 {
+    /// <summary>
+    /// Extension methods for configuring a <see cref="BlobServiceClient"/> in an <see cref="IServiceCollection"/>.
+    /// </summary>
     public static class BlobClientRegistrationExtensions
     {
-        public const string BlobStoreConfigurationSectionName = "BlobStore";
-
+        [Obsolete("Please use " + nameof(AddBlobServiceClient) + " with " + nameof(BlobClientBuilderExtensions.InitializeContainer) + " instead.")]
         public static IServiceCollection AddBlobDataStore(this IServiceCollection services, Action<BlobDataStoreConfiguration> configureAction = null)
```
smithago, line 24: I am assuming there is no health check in azure functions.

So you need to do everything that is here and skip BlobClientProvider and ...TestProvider.
+Add(blobClientInitializer.CreateBlobClient(blobDataStoreConfiguration));
+Add(blobClientInitializer.InitializeDataStoreAsync(_blobServiceClient, blobDataStoreConfiguration, collectionInitializers));

src/Microsoft.Health.Blob/Registration/DefaultBlobDataStoreConfiguration.cs
wsugarman: This is one of the hooks I was reading up on to configure options when they're resolved. I've removed this class and instead used a helpful utility method instead, but the logic is the same. I configure a delegate to modify the options every time they're resolved. So for `IOptions<T>` it happens once, but for `IOptionsSnapshot` this will happen every request as the service has a `Scoped` lifetime.

These configurations shouldn't mess with how the options extensions lib resolves options. In fact, the underlying factory that creates the options looks for these configurations. I was reading about it here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-5.0#options-interfaces

pengchen0692 (approved):

smithago (approved):

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
wsugarman: Yeah I've refactored this

src/Microsoft.Health.Blob/Registration/DefaultBlobDataStoreConfiguration.cs
wsugarman: This class has been removed and its logic has been moved to a `PostConfigure` method invocation

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
wsugarman: They'll be executed sequentially based on the order they were added to the service collection

PR #265, opened 2021-07-08
Add NoOpProcessTerminator, add config to specify which terminator to …

YazanMSFT:
…register. Default will be NoOp

## Description
We should register the NoOpProcessTerminator by default. A config was added to reigster ProcessTerminator when desired.

## Related issues
Addresses [issue #].

## Testing
Tested using the sql test web project.

### DIFF
--- a/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
@@ -35,8 +35,14 @@ namespace Microsoft.Health.SqlServer.Configs
         public SqlServerSchemaOptions SchemaOptions { get; set; } = new SqlServerSchemaOptions();
 
         /// <summary>
-        /// If set, Instructs the serivce to terminate when its schema reaches the specified version.
+        /// If set, Instructs the service to terminate when its schema reaches the specified version.
         /// </summary>
         public int? TerminateWhenSchemaVersionUpdatedTo { get; set; }
+
+        /// <summary>
+        /// true: registers ProcessTerminator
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
```diff
@@ -35,8 +35,14 @@ public class SqlServerDataStoreConfiguration
...
-        /// If set, Instructs the serivce to terminate when its schema reaches the specified version.
+        /// If set, Instructs the service to terminate when its schema reaches the specified version.
         /// </summary>
         public int? TerminateWhenSchemaVersionUpdatedTo { get; set; }
+
+        /// <summary>
+        /// true: registers ProcessTerminator
+        /// false: registers NoOpProcessTerminator
+        /// </summary>
+        public bool RegisterProcessTerminator { get; set; }
```
Ivanidzo4ka, line 46: >RegisterProcessTerminator [](http://example.com/codeflow?start=20&length=25)

wouldn't it be simplier to check TerminateWhenSchemaVersionUpdatedTo and register Noop if it's null, and normal one in other cases?

It feels weird to have two settings. It's makes posible to have noop and schema to be not null, and do nothing.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
YazanMSFT: Agreed, removed the new setting.

PR #270, opened 2021-07-15
Add uniqueidentifier Column Support

wsugarman:
## Description
- Add `UniqueIdentifierColumn` and `NullableUniqueIdentifierColumn` to better support code-gen for `uniqueidentifier`

## Related issues
[AB#83611](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83611)

## Testing
Pass unit tests

Feature

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
@@ -301,6 +301,25 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Model
         }
     }
 
+    public class UniqueIdentifierColumn : Column<Guid>
+    {
+        public UniqueIdentifierColumn(string name)
+            : base(name, SqlDbType.UniqueIdentifier, false)
+        {
+        }
+
+        public override Guid Read(SqlDataReader reader, int ordinal)
+        {
+            return reader.GetGuid(Metadata.Name, ordinal);
+        }
+
+        public override void Set(SqlDataRecord record, int ordinal, Guid value)
+        {
+            EnsureArg.IsNotNull(record, nameof(record));
+            record.SetGuid(ordinal, value);
+        }
+    }
+
     public class NullableIntColumn : Column<int?>
     {
         public NullableIntColumn(string name)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -550,6 +569,33 @@ public override void Set(SqlDataRecord record, int ordinal, byte? value)
         }
     }
 
+    public class NullableUniqueIdentifierColumn : Column<Guid?>
```
brendankowitz, line 572: Quick question: does this also mean only 1 row can have the 'null' identifier. Does this make sense?

wsugarman: I'm sorry, could you rephrase your question? Like the other nullable column types in this file, I'm pretty sure they refer to a representation of some table column type that has been annotated as nullable.

For example, given the SQL table definition:
```SQL
CREATE TABLE Foo
(
    -- ...
    Bar INT NULL,
    -- ...
)
```

In C# we'll code-gen the corresponding field for its table type:
```csharp
internal readonly NullableIntColumn Bar = new NullableIntColumn("Bar");
```

But multiple columns can be annotated as "nullable" for a given row, and multiple rows can contain `NULL` values in these "nullable" columns.

brendankowitz: Initially I was thinking there was also a uniqueness constraint here, where in re-reading, this change for `UniqueIdentifier` is just the datatype. Sorry for the confusion :)

brendankowitz (approved):
:shipit:

src/Microsoft.Health.SqlServer/Features/Schema/Model/Column.cs
wsugarman: Ah no problem! Yeah -- it seems that we're missing support for the type `uniqueidentifier`, so this PR is adding it

wsugarman:
I have written the `[Nullable]UniqueIdentifierColumn` types to operate on the C# type `Guid` instead of `SqlGuid`. It seems we're already using `Guid` when generating stored procedures with parameters of type `uniqueidentifier`, so I think this change is consistent. The only difference is in how these type perform their comparison operations, detailed [here](https://docs.microsoft.com/en-us/sql/connect/ado-net/sql/compare-guid-uniqueidentifier-values?view=sql-server-ver15).

PR #272, opened 2021-07-20
Allow specifying an app id for authenticating using MI to SQL server and Blob storage

YazanMSFT:
## Description
Allow specifying an app id for authenticating using MI to SQL server and Blob storage

## Related issues
Addresses [issue #].

## Testing
Tested by deploying a personal DICOM oss instance

### DIFF
--- a/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Health.Blob.Configs
         public BlobDataStoreRequestOptions RequestOptions { get; } = new BlobDataStoreRequestOptions();
 
         /// <summary>
-        /// If set, the managed identity client ID of the user assigned managed identity to use when connecting to azure storage, if AuthenticationType == ManagedIdentity.
+        /// If set, the client id of the managed identity to use when connecting to azure storage, if AuthenticationType == ManagedIdentity.
         /// </summary>
         public string ManagedIdentityClientId { get; set; }
     }
--- a/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
@@ -14,5 +14,10 @@ namespace Microsoft.Health.Blob.Configs
         public BlobDataStoreAuthenticationType AuthenticationType { get; set; } = BlobDataStoreAuthenticationType.ConnectionString;
 
         public BlobDataStoreRequestOptions RequestOptions { get; } = new BlobDataStoreRequestOptions();
+
+        /// <summary>
+        /// If set, the app ID of the user assigned managed identity to use when connecting to azure storage, if AUthenticationType == ManagedIdentity.
+        /// </summary>
+        public string UserAssignedManagedIdentityAppId { get; set; }
     }
 }
--- a/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
@@ -16,8 +16,8 @@ namespace Microsoft.Health.Blob.Configs
         public BlobDataStoreRequestOptions RequestOptions { get; } = new BlobDataStoreRequestOptions();
 
         /// <summary>
-        /// If set, the app ID of the user assigned managed identity to use when connecting to azure storage, if AUthenticationType == ManagedIdentity.
+        /// If set, the managed identity client ID of the user assigned managed identity to use when connecting to azure storage, if AuthenticationType == ManagedIdentity.
         /// </summary>
-        public string UserAssignedManagedIdentityAppId { get; set; }
+        public string ManagedIdentityClientId { get; set; }
     }
 }
--- a/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
+++ b/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
@@ -26,8 +26,8 @@ namespace Microsoft.Health.Blob.Features.Storage
 
             if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
             {
-                var defaultCredentials = new DefaultAzureCredential();
-                return new BlobServiceClient(new Uri(configuration.ConnectionString), defaultCredentials, blobClientOptions);
+                var managedIdentityCredential = new ManagedIdentityCredential(clientId: configuration.UserAssignedManagedIdentityAppId);
+                return new BlobServiceClient(new Uri(configuration.ConnectionString), managedIdentityCredential, blobClientOptions);
             }
 
             return new BlobServiceClient(configuration.ConnectionString, blobClientOptions);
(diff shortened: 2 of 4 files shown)


src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
```diff
@@ -26,8 +26,8 @@ public static BlobServiceClient Create(BlobDataStoreConfiguration configuration)
 
             if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
             {
-                var defaultCredentials = new DefaultAzureCredential();
-                return new BlobServiceClient(new Uri(configuration.ConnectionString), defaultCredentials, blobClientOptions);
+                var managedIdentityCredential = new ManagedIdentityCredential(clientId: configuration.UserAssignedManagedIdentityAppId);
```
rbans96, line 29: Just for my understanding, is there any possibility that configuration.UserAssignedManagedIdentityAppId is null and we want to use defaultCredentials for ManagedIdentity?

src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
```diff
@@ -14,5 +14,10 @@ public class BlobDataStoreConfiguration
         public BlobDataStoreAuthenticationType AuthenticationType { get; set; } = BlobDataStoreAuthenticationType.ConnectionString;
 
         public BlobDataStoreRequestOptions RequestOptions { get; } = new BlobDataStoreRequestOptions();
+
+        /// <summary>
+        /// If set, the app ID of the user assigned managed identity to use when connecting to azure storage, if AUthenticationType == ManagedIdentity.
+        /// </summary>
+        public string UserAssignedManagedIdentityAppId { get; set; }
```
brandonpollett, line 21: Does it have to be a user assigned managed identity, or do system assigned also work here? Additionally, should we call this `ManagedIdentityClientId` to map with the parameter for the `ManagedIdentityCredential` type?

src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
```diff
@@ -14,5 +14,10 @@ public class BlobDataStoreConfiguration
         public BlobDataStoreAuthenticationType AuthenticationType { get; set; } = BlobDataStoreAuthenticationType.ConnectionString;
 
         public BlobDataStoreRequestOptions RequestOptions { get; } = new BlobDataStoreRequestOptions();
+
+        /// <summary>
+        /// If set, the app ID of the user assigned managed identity to use when connecting to azure storage, if AUthenticationType == ManagedIdentity.
```
brandonpollett, line 19: nit: `AUthenticationType` => `AuthenticationType`

src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
brandonpollett: I would say that is a valid use case. I would think this is more of an override.

src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
brandonpollett: Interesting the naming inconsistency between SQL and `ManagedIdentityCredential`.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -81,7 +81,19 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
 
             // The following are only used in case of managed identity
             services.AddSingleton<IAccessTokenHandler, ManagedIdentityAccessTokenHandler>();
-            services.AddSingleton<AzureServiceTokenProvider>();
+            services.AddSingleton<AzureServiceTokenProvider>(p =>
```
brandonpollett, line 84: Not sure it's any easier, but it appears that `Microsoft.Data.SqlClient` has some support for just saying use managed identity in the connection string. https://docs.microsoft.com/en-us/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver15#using-active-directory-managed-identity-authentication

src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
YazanMSFT: The parameter documentation in the SDK explicitly says user assigned:

// Parameters:
        //   clientId:
        //     The client id to authenticate for a user assigned managed identity. More information
        //     on user assigned managed identities can be found here: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-a-user-assigned-managed-identity-works-with-an-azure-vm
        //

YazanMSFT: It defaults to null so not specifying it should switch back to system assigned.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
YazanMSFT: I think this will require reprovisioning all existing accounts

src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
YazanMSFT: When it's null it should revert to using a system assigned managed identity. DefaultCredentials will try other things like VS credentials, environment credentials, etc. I can make it use DefaultCredentials if ManagedIdentity is null or empty, if that's what we want.

brandonpollett: I would say we should probably fallback to using `DefaultAzureCredential` if the `ManagedIdentityClientId` is null or empty. My main reason is that otherwise this is a breaking change. 

Would it be easier to simply specify the clientId in the options for `DefaultAzureCredential` like in this example? https://docs.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet#examples

YazanMSFT: Makes sense. Yeah that looks easier, let me give it a try.

src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
```diff
@@ -46,5 +46,10 @@ public class SqlServerDataStoreConfiguration
         /// If set, Instructs the service to terminate when its schema reaches the specified version.
         /// </summary>
         public int? TerminateWhenSchemaVersionUpdatedTo { get; set; }
+
+        /// <summary>
+        /// If set, the app ID of the user assigned managed identity to use when connecting to SQL, if AUthenticationType == ManagedIdentity.
+        /// </summary>
+        public string UserAssignedManagedIdentityAppId { get; set; }
```
brandonpollett, line 53: Do you think it makes sense to make this align with the other setting? Not sure it's a huge deal. 

Also, `AUthenticationType` -> `AuthenticationType`

brandonpollett (approved):

rbans96 (approved):

PR #273, opened 2021-07-21
Update the Schema Status to lowercase to be consistent with the statu…

apurvabhaleMS:
## Description
Update the Schema Status to lowercase to be consistent with the statuses applied when AutomaticUpdates is turned on

## Related issues
Addresses [issue [81910](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81910)].

## Testing
Manual Testing

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaVersionStatus.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaVersionStatus.cs
@@ -11,8 +11,16 @@ namespace Microsoft.Health.SqlServer.Features.Schema
     [JsonConverter(typeof(StringEnumConverter))]
     public enum SchemaVersionStatus
     {
-        Started,
-        Completed,
-        Failed,
+        // Note - Updating SchemaVersionStatus to lower case to be consistent with the statuses applied when schema AutomaticUpdates is set to true
+
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+        started,
+#pragma warning restore SA1300 // Element should begin with upper-case letter
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+        completed,
+#pragma warning restore SA1300 // Element should begin with upper-case letter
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/SchemaVersionStatus.cs
```diff
@@ -11,8 +11,16 @@ namespace Microsoft.Health.SqlServer.Features.Schema
     [JsonConverter(typeof(StringEnumConverter))]
     public enum SchemaVersionStatus
     {
-        Started,
-        Completed,
-        Failed,
+        // Note - Updating SchemaVersionStatus to lower case to be consistent with the statuses applied when schema AutomaticUpdates is set to true
```
rbans96, line 14: nit: schema AutomaticUpdates -> SchemaOptions:AutomaticUpdatesEnabled?

src/Microsoft.Health.SqlServer/Features/Schema/SchemaVersionStatus.cs
```diff
@@ -11,8 +11,16 @@ namespace Microsoft.Health.SqlServer.Features.Schema
...
+
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+        started,
+#pragma warning restore SA1300 // Element should begin with upper-case letter
```
rbans96, line 18: I am wondering, if we could have warning disable at the top and warning restore at the bottom and avoid the intermediate ones?

apurvabhaleMS: just updated

src/Microsoft.Health.SqlServer/Features/Schema/SchemaVersionStatus.cs
apurvabhaleMS: Yep! Added the exact name

rbans96 (approved):

PR #279, opened 2021-07-27
Schema initialization/migration schemaversion status issue

apurvabhaleMS:
## Description
Delete an entry from SchemaVersion table with 'failed' status before applying new schema

For Testing -

1. Set the SchemaVersionConstants.Max to 15 in FHIR server
2. Set "SqlServer:SchemaOptions:AutomaticUpdatesEnabled": "true"
3. To generate the error -> Update 15.sql to Insert the 'started' status in SchemaVersion table right before commit transaction -> this will result into an error and everything will be rolled back
4. Now run the server
5. In the DB you should see an entry with 'failed' status in SchemaVersion table for version 15
6. Revert the step 3.
7. Now run the server again
8. You will notice that SchemaVersion's 15 - failed entry will be deleted
9. Schema will be created and SchemaVersion will have 15 - completed row

## Related issues
Addresses [[81613](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81613)].

## Testing
Manual Testing
Unit Testing

Patch|Skip|Feature|Breaking

### DIFF
--- a/test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaUpgradeRunnerTests.cs
+++ b/test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaUpgradeRunnerTests.cs
@@ -62,14 +62,35 @@ namespace Microsoft.Health.SqlServer.Tests.Integration.Features.Schema
         }
 
         [Fact]
-        public async Task ApplySchemaOnPreviouslyFailedAttempt_UsingDiff_Succeeds()
+        public async Task ApplySchema_UsingDiff_Succeeds()
         {
             await _runner.ApplyBaseSchemaAsync(CancellationToken.None);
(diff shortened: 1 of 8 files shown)


rbans96:
Can we add unit test for this usecase?

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/SchemaUpgradeRunnerTests.cs
```diff
@@ -62,9 +62,11 @@ public async Task ApplySchema_BaseSchemaExists_Succeeds()
         }
 
         [Fact]
-        public async Task ApplySchema_UsingDiff_Succeeds()
+        public async Task ApplySchemaOnPreviouslyFailedAttempt_UsingDiff_Succeeds()
```
rbans96, line 65: Should not we add another test and keep the existing one because they are two different usecases?

rbans96 (approved):

PR #281, opened 2021-07-31
Supporting optional CancellationToken parameters in blob storage library

jovinson-ms:
## Description
Adds optional `CancellationToken` support throughout the blob storage library. The main scenario this affects is the host cancelling `BlobHostedService` as part of the hosted service lifecycle - cancellation is propagated to every blob client call that can be cancelled.

Exceptions were being silently handled in `BlobHealthCheck`, but now exceptions are thrown, as this is the [recommended pattern.](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-5.0#create-health-checks).

Also added / cleaned up some documentation.

Small note: the cancellation token library uses the double `l` spelling throughout, but throws a single `l` `OperationCanceledException`. Apparently, [both are acceptable.](https://www.dictionary.com/e/canceled-vs-cancelled/) 

## Testing
This change was tested against existing unit tests. Unit tests were added to test cancellation handling.

### DIFF
--- a/src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
+++ b/src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
@@ -12,6 +12,7 @@ using Azure.Storage.Blobs.Models;
 using Azure.Storage.Blobs.Specialized;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Health.Blob.Features.Storage;
+using Microsoft.IO;
 using NSubstitute;
 using Xunit;
 
@@ -34,18 +35,18 @@ namespace Microsoft.Health.Blob.UnitTests.Features.Storage
                     return Substitute.For<Response<BlobContentInfo>>();
                 });
 
-            var blobContainerClient1 = Substitute.For<BlobContainerClient>(new Uri("https://www.microsoft.com/"), new BlobClientOptions());
-            blobContainerClient1.GetBlockBlobClient(Arg.Any<string>()).Returns(blockBlobClient);
+            var blobContainerClient = Substitute.For<BlobContainerClient>(new Uri("https://www.microsoft.com/"), new BlobClientOptions());
+            blobContainerClient.GetBlockBlobClient(Arg.Any<string>()).Returns(blockBlobClient);
 
             _blobClient = Substitute.For<BlobServiceClient>(new Uri("https://www.microsoft.com/"), null);
-            _blobClient.GetBlobContainerClient(Arg.Any<string>()).Returns(blobContainerClient1);
+            _blobClient.GetBlobContainerClient(Arg.Any<string>()).Returns(blobContainerClient);
         }
 
         [Fact]
         public async void GivenCancellation_WhenPerformingTest_ThenExceptionIsHandled()
         {
-            var testProvider = new BlobClientReadWriteTestProvider(new IO.RecyclableMemoryStreamManager(), _logger);
-            var cancellationTokenSource = new CancellationTokenSource();
+            var testProvider = new BlobClientReadWriteTestProvider(new RecyclableMemoryStreamManager(), _logger);
+            using var cancellationTokenSource = new CancellationTokenSource();
             cancellationTokenSource.Cancel();
 
             var exception = await Record.ExceptionAsync(() => testProvider.PerformTestAsync(_blobClient, new Configs.BlobContainerConfiguration(), cancellationTokenSource.Token));
--- a/src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
+++ b/src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
@@ -1,0 +1,55 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Threading;
+using Azure;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using Azure.Storage.Blobs.Specialized;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Health.Blob.Features.Storage;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Blob.UnitTests.Features.Storage
+{
+    public class BlobClientReadWriteTestProviderTests
+    {
+        private readonly NullLogger<BlobClientReadWriteTestProvider> _logger;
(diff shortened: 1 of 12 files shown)


jovinson-ms:
Addresses AB#83994

src/Microsoft.Health.Blob/Features/Storage/BlobClientReadWriteTestProvider.cs
```diff
@@ -37,12 +46,20 @@ public async Task PerformTestAsync(BlobServiceClient client, BlobContainerConfig
...
+                using var content = new MemoryStream(Encoding.UTF8.GetBytes(TestBlobContent));
+                await blob.UploadAsync(
+                    content,
+                    new BlobHttpHeaders { ContentType = "text/plain" },
+                    cancellationToken: cancellationToken).ConfigureAwait(false);
+                await DownloadBlobContentAsync(blob, cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException ex)
+            {
+                _logger.LogCritical(ex, "Reading and writing cancelled for blob: {container}/{blob}", blobContainerConfiguration.ContainerName, TestBlobName);
```
wsugarman, line 61: Should we instead move this logging to the top of the call stack? Unfortunately I don't know where this is called. I am unsure if we invoke this method. Perhaps I am just blind lol

src/Microsoft.Health.Blob/Features/Storage/BlobContainerInitializer.cs
```diff
@@ -25,14 +30,22 @@ public BlobContainerInitializer(string containerName, ILogger<BlobContainerIniti
...
-            await container.CreateIfNotExistsAsync();
+            _logger.LogDebug("Creating blob container if not exists: {containerName}", _containerName);
+
+            try
+            {
+                await container.CreateIfNotExistsAsync(cancellationToken: cancellationToken);
+            }
+            catch (OperationCanceledException ex)
+            {
+                _logger.LogCritical(ex, "Blob container initialization cancelled: {containerName}", _containerName);
```
wsugarman, line 47: I think you meant to include a `throw;` here

wsugarman, line 47: Same comment as the other file. Should we log this or simply log at the top of the call stack in the hosted service.

src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            var blockBlobClient = Substitute.For<BlockBlobClient>();
+            blockBlobClient.UploadAsync(Arg.Any<MemoryStream>(), Arg.Any<BlobHttpHeaders>(), null, null, null, null, Arg.Any<CancellationToken>())
+                .Returns(x =>
+                {
+                    x.Arg<CancellationToken>().ThrowIfCancellationRequested();
+                    return Substitute.For<Response<BlobContentInfo>>();
+                });
+
+            var blobContainerClient1 = Substitute.For<BlobContainerClient>(new Uri("https://www.microsoft.com/"), new BlobClientOptions());
+            blobContainerClient1.GetBlockBlobClient(Arg.Any<string>()).Returns(blockBlobClient);
```
wsugarman, line 38: nit: why is this "1"?

src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            _blobClient = Substitute.For<BlobServiceClient>(new Uri("https://www.microsoft.com/"), null);
+            _blobClient.GetBlobContainerClient(Arg.Any<string>()).Returns(blobContainerClient1);
+        }
+
+        [Fact]
+        public async void GivenCancellation_WhenPerformingTest_ThenExceptionIsHandled()
+        {
+            var testProvider = new BlobClientReadWriteTestProvider(new IO.RecyclableMemoryStreamManager(), _logger);
+            var cancellationTokenSource = new CancellationTokenSource();
```
wsugarman, line 48: nit: use `using` statement. You can use the [`using declaration`](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations)

src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobContainerInitializerTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            _blobClient = Substitute.For<BlobServiceClient>(new Uri("https://www.microsoft.com/"), null);
+            _blobClient.GetBlobContainerClient(TestContainerName1).Returns(blobContainerClient1);
+        }
+
+        [Fact]
+        public async void GivenCancellation_WhenInitializingContainer_ThenExceptionIsHandled()
+        {
+            var blobContainerInitializer = new BlobContainerInitializer(TestContainerName1, _logger);
+            var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromMilliseconds(1));
```
wsugarman, line 45: nit: use `using` statement. You can use the [`using declaration`](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations)

src/Microsoft.Health.Blob/Features/Storage/BlobClientReadWriteTestProvider.cs
wsugarman, line 61: Missing `throw;` statement here if we keep the logging

src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
jovinson-ms: copy / paste artifact - removing.

src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobContainerInitializerTests.cs
jovinson-ms: done

src/Microsoft.Health.Blob.UnitTests/Features/Storage/BlobClientReadWriteTestProviderTests.cs
jovinson-ms: done

src/Microsoft.Health.Blob/Features/Storage/BlobClientReadWriteTestProvider.cs
jovinson-ms: fixed

src/Microsoft.Health.Blob/Features/Storage/BlobContainerInitializer.cs
jovinson-ms: fixed

wsugarman (approved):

Issue #282, opened 2021-08-01
Set value for varbinary column would not clean old value


tongwu-msft:
**Describe the bug**
Set value for varbinary column would not clean old value, if the old value length is longer than new value there would be invalid bytes after new value. For the method "ToDataRecordEnumerable", the SqlDataRecord is used by multiple rows, and varbinary value is wrong.

**To Reproduce**
1. TableValuedParameterDefinition add parameters with 2 rows contains varbinary column
2. Row 1 varbinary: 0x1111, row 2 varbinary: 0x22
3. after add parameters, the 2nd row varbinary value become 0x2211

**Expected behavior**
Row 2 should have correct value and length short than row 1

**Actual behavior**
Row 2 value become 0x2211

PR #283, opened 2021-08-01
Clear old value before setting varbinary column

tongwu-msft:
## Description
Clear old value before setting varbinary column to fix the issue #282 

## Related issues
Addresses [issue #282 ].

## Testing

Patch|Skip|Feature|Breaking


brendankowitz (approved):

brendankowitz:
If possible, add a regression test for this, other than that LGTM

tongwu-msft:
> If possible, add a regression test for this, other than that LGTM
Added.

PR #284, opened 2021-08-02
Allow register specific module to support DicomFunctions

pengchen0692:
## Description
DicomFunctions  only need register small subset of modules in assembly, add RegistryModule to support it.

## Related issues
Addresses [AB#84056](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/84056)

## Testing
* All existing automated tests pass.

### DIFF
--- a/src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -24,11 +24,23 @@ namespace Microsoft.Health.Extensions.DependencyInjection
         /// <param name="constructorParams">An array of constructor parameters that can be injected into IStartupModule on creation.</param>
         public static void RegisterAssemblyModules(this IServiceCollection collection, Assembly assembly, params object[] constructorParams)
         {
+            collection.RegisterAssemblyModules(assembly: assembly, moduleTypeFilter: null, constructorParams: constructorParams);
+        }
+
+        /// <summary>
+        /// Registers assembly modules.
+        /// </summary>
+        /// <param name="collection">The collection.</param>
+        /// <param name="assembly">The assembly.</param>
+        /// <param name="moduleTypeFilter">Filter out modules that are not be registered.</param>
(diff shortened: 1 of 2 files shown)


jovinson-ms (approved):

src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -23,12 +23,24 @@ public static class ServiceCollectionExtensions
...
+        }
+
+        /// <summary>
+        /// Registers assembly modules.
+        /// </summary>
+        /// <param name="collection">The collection.</param>
+        /// <param name="assembly">The assembly.</param>
+        /// <param name="moduleTypeFilter">Filter out modules that are not be registered.</param>
+        /// <param name="constructorParams">An array of constructor parameters that can be injected into IStartupModule on creation.</param>
+        public static void RegisterAssemblyModules(this IServiceCollection collection, Assembly assembly, Func<Type, bool> moduleTypeFilter, params object[] constructorParams)
```
brendankowitz, line 37: nit:
```suggestion
        public static void RegisterAssemblyModules(this IServiceCollection collection, Assembly assembly, Predicate<Type> moduleTypeFilter, params object[] constructorParams)
```

<br />
Another thought, from the description of your use-case, did you consider something like this instead?

`public static void RegisterModule(this IServiceCollection collection, IStartupModule module)`?

wsugarman, line 37: Another alternative: could we instead simply allow users to register specific modules, like:
```suggestion
        public static void RegisterModule<T>(this IServiceCollection collection, params object[] constructorParams) where T : IStartupModule
```

pengchen0692: @brendankowitz  thanks Brendan. I'm thinking to organize modules into different namespaces, one for DicomFunctions, the other for DicomWebServer. Would that make sense to you?

wsugarman (approved):

PR #301, opened 2021-08-24
Handle exceptions when schema does not exist

apurvabhaleMS:
## Description
When Schema is not initialized (SchemaOptions:AutomaticUpdatesEnabled : false), SchemaJob, ExportJob, Task hosting background services throw errors as they try to access stored procedure which do not exist. With this PR we are trying to avoid multiple exception loggings over and over by checking if schemaInformation.current version in null (implying Schema is not initialized). Instead we are logging warnings for the same issues.

## Related issues
Addresses [[81886](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81886)].

## Testing
Manual Testing -

1. Set "SqlServer:SchemaOptions:AutomaticUpdatesEnabled" as false in launchSettings.cs.
 2. Run Fhir server
 3. Notice there are no errors are logged in console
 4. You should see warning loggings for Could not find stored procedure...

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
@@ -87,7 +87,7 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Manager
                 object current = await selectCommand.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false);
                 return (current == null || Convert.IsDBNull(current)) ? 0 : (int)current;
             }
-            catch (SqlException e) when (string.Equals(e.Message, Resources.CurrentSchemaVersionStoredProcedureNotFound, StringComparison.OrdinalIgnoreCase))
+            catch (SqlException e) when (string.Equals(e.Message, string.Format(Resources.CurrentSchemaVersionStoredProcedureNotFound, "dbo.SelectCurrentSchemaVersion"), StringComparison.OrdinalIgnoreCase))
             {
                 return 0;
             }
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
@@ -8,6 +8,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using MediatR;
+using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -81,6 +82,10 @@ namespace Microsoft.Health.SqlServer.Features.Schema
                     {
                         _processTerminator.Terminate(cancellationToken);
                     }
+                }
+                catch (SqlException e) when (schemaInformation.Current is null && string.Equals(e.Message, string.Format(Resources.CurrentSchemaVersionStoredProcedureNotFound, "dbo.UpsertInstanceSchema"), StringComparison.OrdinalIgnoreCase))
+                {
+                    _logger.LogWarning(e, "Schema is not initialized.");
                 }
                 catch (Exception ex)
                 {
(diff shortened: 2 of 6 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -82,6 +83,10 @@ public async Task ExecuteAsync(SchemaInformation schemaInformation, string insta
                         _processTerminator.Terminate(cancellationToken);
                     }
                 }
+                catch (SqlException e) when (schemaInformation.Current is null && string.Equals(e.Message, string.Format(Resources.CurrentSchemaVersionStoredProcedureNotFound, "dbo.UpsertInstanceSchema"), StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
apurvabhaleMS: For this one, I think we can use SchmaInformation.current to check against

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
```diff
@@ -123,23 +116,29 @@ public async Task ExecuteScriptAsync(string script, CancellationToken cancellati
 
         /// <inheritdoc />
         public async Task<bool> BaseSchemaExistsAsync(CancellationToken cancellationToken)
+        {
+            return await ObjectExistsAsync("SelectCurrentVersionsInformation", "P", cancellationToken);
```
wsugarman, line 120: nit: no need to `await` here. Simply pass the task directly:
```suggestion
            return ObjectExistsAsync("SelectCurrentVersionsInformation", "P", cancellationToken);
```

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
```diff
@@ -123,23 +116,29 @@ public async Task ExecuteScriptAsync(string script, CancellationToken cancellati
...
+            command.Parameters.AddWithValue("@type", objectType);
 
             return (int)await command.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false) != 0;
         }
 
         /// <inheritdoc />
         public async Task<bool> InstanceSchemaRecordExistsAsync(CancellationToken cancellationToken)
         {
-            var procedureQuery = "SELECT COUNT(*) FROM dbo.InstanceSchema";
+            var procedureQuery = "SELECT COUNT(*) FROM dbo.InstanceSchema1";
```
wsugarman, line 141: Was this intentional?

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -113,14 +114,22 @@ public async Task InitializeAsync(bool forceIncrementalSchemaUpgrade = false, Ca
...
             {
-                _logger.LogInformation("No version found. It must be new database");
+                int version = await _schemaManagerDataStore.GetCurrentSchemaVersionAsync(cancellationToken);
+                if (version != 0)
+                {
+                    _schemaInformation.Current = version;
+                }
+                else
+                {
+                    _logger.LogInformation("No version found. It must be new database");
```
wsugarman, line 131: Would we use this same message if `_canCallGetCurrentSchema` is `false`?

wsugarman (approved):

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
apurvabhaleMS: good point. Added a warning for SP not found

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
apurvabhaleMS: ohh reverted

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
apurvabhaleMS: umm, it expects return type as bool rather than  Task<bool>. I guess I am not following :(

PR #303, opened 2021-08-31
Expose AddBlobContainerInitialization and Encourage Usage over Existing APIs

wsugarman:
## Description
Current API for adding blob storage circumvents the dependency injection service container by directly creating a `BlobServiceClient`. Instead, developers should leverage the [`Microsoft.Extensions.Azure` NuGet package](https://www.nuget.org/packages/Microsoft.Extensions.Azure/) to fully integrate and receive more diagnostic information via `ILogger`. To facilitate this change, this PR includes:
- Remove the dependency on `BlobDataStoreConfiguration` outside of initializing the `BlobServiceClient` (if it's done via APIs in this package)
- Expose `AddBlobContainerInitialization` to allow the decoupling of registering a client and adding the hosted service
- Added `BlobServiceClientOptions` to encapsulate all of the data we'll most likely specify in our configurations for our `BlobServiceClient` objects
- Add new `AddBlobServiceClient` API to use `Microsoft.Extensions.Azure`

## Related issues
[AB#84923](https://microsofthealth.visualstudio.com/Health/_workitems/edit/84923)

## Testing
Tested locally and via pipeline

Breaking (due to ctor change)


src/Microsoft.Health.Blob/Configs/BlobServiceClientOptions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Azure.Storage.Blobs;
+
+namespace Microsoft.Health.Blob.Configs
+{
+    public class BlobServiceClientOptions : BlobClientOptions
```
wsugarman, line 10: This type may be unnecessary

jovinson-ms (approved):

PR #306, opened 2021-09-10
Refactor Blob Initialization

wsugarman:
## Description
Move code that adds instances of `BlobContainerInitializer` into the service container into the core libraries using the builder pattern, similar to `OptionsBuilder`.

You can see we repeat this logic in DICOM:
- [BlobStore](https://github.com/microsoft/dicom-server/blob/606b53e250029ad593b858f5c7cc9fac758d12ce/src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs#L62)
- [MetadataStore](https://github.com/microsoft/dicom-server/blob/606b53e250029ad593b858f5c7cc9fac758d12ce/src/Microsoft.Health.Dicom.Metadata/Registration/DicomServerBuilderMetadataRegistrationExtensions.cs#L52)

## Related issues
[AB#84923](https://microsofthealth.visualstudio.com/Health/_workitems/edit/84923)

## Testing
Tested via pipeline

Patch
(Although it is actually breaking, but the previous breaking change 3.2 was unlisted for blob)

### DIFF
--- a/src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
+++ b/src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
@@ -12,9 +12,11 @@ using EnsureThat;
 using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Blob.Configs;
 using Microsoft.Health.Blob.Features.Storage;
+using Microsoft.Health.Blob.Registration;
 using Microsoft.IO;
 
 namespace Microsoft.Extensions.DependencyInjection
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -143,28 +145,70 @@ public static IServiceCollection AddBlobServiceClient(this IServiceCollection se
...
         /// <param name="configure">A delegate for configuring the <see cref="BlobInitializerOptions"/>.</param>
-        /// <returns>The <paramref name="services"/> for additional method invocations.</returns>
+        /// <returns>
+        /// A <see cref="BlobContainerInitializationBuilder"/> for configuring one or more containers.
+        /// </returns>
         /// <exception cref="ArgumentNullException">
         /// <paramref name="services"/> is <see langword="null"/>.
         /// </exception>
-        public static IServiceCollection AddBlobContainerInitialization(this IServiceCollection services, Action<BlobInitializerOptions> configure = null)
+        public static BlobContainerInitializationBuilder AddBlobContainerInitialization(this IServiceCollection services, Action<BlobInitializerOptions> configure = null)
```
wsugarman, line 159: I made a breaking change here, but I'm going to delist my previous version.

jovinson-ms (approved):

PR #314, opened 2021-09-21
Create ByteCounterStream to capture the bytes written to a response

jnlycklama:
## Description
ByteCounterStream can be set as the context.Response.Body, and will then capture all bytes written to the stream.

This code has been copied from health-pass https://microsofthealth.visualstudio.com/Health/_git/health-paas?path=/src/Common/ServicePlatform/ByteCounting&_a=contents&version=GBmaster

To be used by: https://github.com/microsoft/dicom-server/pull/1046

## Related issues
Addresses 85307

## Testing
Unit tests

Skip


src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Api.Features.ByteCounter
+{
+    /// <summary>
+    /// A stream to wrap an underlying stream and counts the number of bytes passed while operating with it.
+    /// </summary>
+    public class ByteCountingStream : Stream
```
wsugarman, line 17: This probably makes sense in the "Core" project rather than API, right?

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// <inheritdoc />
+        public override bool CanWrite => _stream.CanWrite;
+
+        /// <inheritdoc />
+        public override long Length => _stream.Length;
+
+        /// <inheritdoc />
+        public override long Position
+        {
+            get { return _stream.Position; }
```
wsugarman, line 48: nit: can still use expressions here:
```suggestion
            get => _stream.Position;
            set => _stream.Position = value;
```

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,135 @@
...
+        private readonly Stream _stream;
+        private long _writtenByteCount;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ByteCountingStream"/> class.
+        /// </summary>
+        /// <param name="stream">An underlying stream.</param>
+        public ByteCountingStream(Stream stream)
+        {
+            EnsureArg.IsNotNull(stream, nameof(stream));
```
wsugarman, line 28: nit:
```suggestion
            _stream = EnsureArg.IsNotNull(stream, nameof(stream));
```

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,135 @@
...
+        public override void SetLength(long value)
+        {
+            _stream.SetLength(value);
+        }
+
+        /// <inheritdoc />
+        public override void WriteByte(byte value)
+        {
+            _stream.WriteByte(value);
+            Interlocked.Increment(ref _writtenByteCount);
```
wsugarman, line 91: I appreciate the thread-safety, but `Stream` but reading and writing are not guaranteed to be thread-safe. Instead, users should only have at most a single thread reading from and a single thread writing to a `Stream` at a time. Otherwise, they need to synchronize their access. `Stream` even provides a static method for synchronizing called `Stream.Synchronized(Stream)` for helping with this.

So I don't think you need to worry about using `Interlocked` to increment this count.

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
wsugarman, line 17: Perhaps a name more like `MonitoredStream` or `AuditedStream` to mirror `BufferedStream`?

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,135 @@
...
+        public override long Position
+        {
+            get { return _stream.Position; }
+            set { _stream.Position = value; }
+        }
+
+        /// <summary>
+        /// The number of bytes that have been written.
+        /// </summary>
+        public long WrittenByteCount { get => _writtenByteCount; }
```
wsugarman, line 55: Would `WriteCount` be sufficient?

wsugarman, line 55: Should we also count the bytes read for completeness?

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            Interlocked.Add(ref _writtenByteCount, count);
+            return _stream.BeginWrite(buffer, offset, count, callback, state);
+        }
+
+        /// <inheritdoc />
+        protected override void Dispose(bool disposing)
+        {
+            _stream.Dispose();
+            base.Dispose(disposing);
```
wsugarman, line 132: If we seal this class, we could probably simplify this to simply do something like:
```csharp
public void Dispose()
  => _stream.Dipose();
```

src/Microsoft.Health.Api/Features/ByteCounter/HeaderUtility.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Text;
+using EnsureThat;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Health.Api.Features.ByteCounter
+{
+    public static class HeaderUtility
+    {
+        private static readonly Encoding HeaderEncoding = Encoding.UTF8;
```
wsugarman, line 16: Is it guaranteed headers are always UTF-8?

wsugarman (changes requested):

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: I put it in API since it is meant to be used by dicom-server API project and added to the response body (see: https://github.com/microsoft/dicom-server/pull/1046)
But I guess it could be used by any kind of Stream so I can see why the "Core" project makes sense

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Ah thanks. We had talked about thread safety in the other PR but I didn't know this about streams.

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Yeah we could, but it won't be used anywhere. This is to get the egress so I don't know if getting the ingress would ever be necessary

src/Microsoft.Health.Api/Features/ByteCounter/IByteCountingStreamFactory.cs
```diff
@@ -0,0 +1,19 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.IO;
+
+namespace Microsoft.Health.Api.Features.ByteCounter
+{
+    public interface IByteCountingStreamFactory
```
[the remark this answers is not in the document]
jnlycklama: Just to keep with the factory pattern. Also copied from fhir code so I was keeping the usage similar.
It doesn't really do anything though so I'd be ok with getting rid of it

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Moved stream to Core project. The HeaderUtility had to stay in Api project since it referenced AspNet, so I made it an extension instead

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Updated

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: I think AuditedStream could get too easily confused with the Audit logging / audit feature. But I like the name MonitoredStream

src/Microsoft.Health.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Even in a sealed class, removing "base.Dispose(disposing);" will pop up an error complaining that base.Dispose(disposing) is missing. I don't think it's worth it to suppress the error so I left it as is

src/Microsoft.Health.Api/Features/ByteCounter/HeaderUtility.cs
jnlycklama: I did a bit of research and it seems this should work for most cases, but I update the HeaderUtility so it specifically asks for an encoding. That way it will be explicit if someone else uses it with a special case

src/Microsoft.Health.Api/Extensions/HeaderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+            EnsureArg.IsNotNull(headerEncoding, nameof(headerEncoding));
+
+            int headerDelimiterByteCount = headerEncoding.GetByteCount(": ");
+            int headerEndOfLineCharactersByteCount = headerEncoding.GetByteCount("\r\n");
+            int headerLength = 0;
+            foreach (KeyValuePair<string, StringValues> header in headers)
+            {
+                headerLength += headerEncoding.GetByteCount(header.Key)
+                    + headerDelimiterByteCount
+                    + headerEncoding.GetByteCount(header.Value.ToString())
```
wsugarman, line 31: `StringValues.ToString()` returns comma-separated values where the commas weren't actually present in the input. However, this may be redoing logic for parsing headers. So I suppose my question is: how are `StringValues` instances created for a given header?
- If it splits comma-separated values in the header (such that `A=1,2,3` is the same as `new StringValues( new List<string> {  "1", "2", "3" })`), then we're probably ok as `ToString` is undoing this logic. But this seems unlikely, as I think different headers are formatted differently, and I can't imagine each header has its own split logic.
- If it is aggregating duplicate headers (such that headers `A=1,2` and `A=3` is the same as `new StringValues( new List<string> {  "1,2", "3" })`), then we need to count the size of each entry in `StringValues` individually

wsugarman (approved):

src/Microsoft.Health.Api/Extensions/HeaderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Text;
+using EnsureThat;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Health.Api.Extensions
+{
+    public static class HeaderExtensions
+    {
+        public static int GetTotalHeaderLength(this IHeaderDictionary headers, Encoding headerEncoding)
```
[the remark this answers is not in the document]
jnlycklama: Thanks for finding that, I looked around but had a hard time finding a concrete answer. Updated

src/Microsoft.Health.Api/Extensions/HeaderExtensions.cs
jnlycklama: Ah good catch! I did a bit of research but couldn't find anything specific on how it works. Doing it the second way should work for all scenarios so I've updated it.

Issue #317, opened 2021-09-27
We should avoid creating `HttpClient`


jovinson-ms:
We should consider injecting `IHttpClientFactory` instead: https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

https://github.com/microsoft/healthcare-shared-components/blob/1e0dc8532113a5f7fdd0692c922e91b8273c240e/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs#L25

wbalbo:
Please assign this to me. Probably will not need a unit test for this, since the constructor is not called anywhere in the code, is just used to instantiate the HttpClient class.

Thanks.

jovinson-ms:
Great! You can remove the `IDisposable` implementation once we we're using the factory, and the existing tests should continue to pass.

wbalbo:
Yes, I had a problem only with tests related to the SQL Server database, but is not because I changed the implementation, is only because I didn't create a SQL connection.

PR #318, opened 2021-09-29
Created an injection dependency of the HttpClient class in the SchemaClient class constructor

wbalbo:
## Description
- Removed IDisposable inheritance
- Changed the constructor implementation to add an injection dependency HttpClient class, to avoid errors using a new instance of this class in every call

## Related issues
Addresses [#317].

## Testing
I ran tests on the SqlSchemaManagerTests class, which uses the ISchemaClient interface, and I didn't find any errors.

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
@@ -20,9 +20,9 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Manager
     {
         private readonly HttpClient _httpClient;
 
-        public SchemaClient(HttpClient httpClient)
+        public SchemaClient(IHttpClientFactory httpClientFactory)
         {
-            _httpClient = httpClient;
+            _httpClient = httpClientFactory.CreateClient();
         }
 
         public void SetUri(Uri uri)
@@ -118,20 +118,6 @@ namespace Microsoft.Health.SqlServer.Features.Schema.Manager
             else
             {
                 throw new SchemaManagerException(string.Format(CultureInfo.InvariantCulture, Resources.ScriptNotFound, response.StatusCode));
-            }
-        }
-
-        public void Dispose()
-        {
-            Dispose(true);
-            GC.SuppressFinalize(this);
-        }
-
-        protected virtual void Dispose(bool disposing)
-        {
-            if (disposing)
-            {
-                _httpClient?.Dispose();
             }
         }
     }
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
@@ -16,13 +16,13 @@ using Newtonsoft.Json;
 
 namespace Microsoft.Health.SqlServer.Features.Schema.Manager
 {
-    public class SchemaClient : ISchemaClient, IDisposable
+    public class SchemaClient : ISchemaClient
     {
         private readonly HttpClient _httpClient;
 
-        public SchemaClient()
+        public SchemaClient(HttpClient httpClient)
         {
-            _httpClient = new HttpClient();
+            _httpClient = httpClient;
         }
 
         public void SetUri(Uri uri)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -16,13 +16,13 @@
 
 namespace Microsoft.Health.SqlServer.Features.Schema.Manager
 {
-    public class SchemaClient : ISchemaClient, IDisposable
+    public class SchemaClient : ISchemaClient
     {
         private readonly HttpClient _httpClient;
 
-        public SchemaClient()
+        public SchemaClient(HttpClient httpClient)
```
jovinson-ms, line 23: Rather than injecting an `HttpClient` directly, let's inject an `IHttpClientFactory` which we can use to access an `HttpClient` efficiently: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-5.0

wbalbo: Hmm, that makes sense.

Sorry, I didn't notice this, you are right.

Thanks.

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -16,13 +16,13 @@
 
 namespace Microsoft.Health.SqlServer.Features.Schema.Manager
 {
-    public class SchemaClient : ISchemaClient, IDisposable
+    public class SchemaClient : ISchemaClient
```
jovinson-ms, line 19: Since we're removing the `IDisposable` interface, let's remove the methods that provide it's implementation on lines 124 - 136.

wbalbo: Yes, this is good advice. Since the code didn't return an error and the tests were working seems to me that is okay. But I forgot to see the disposable implementations.

Thank you.

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
wbalbo: hmm, I needed to install Microsoft.Extensions.Http version 3.1.16 from NuGet. I tried to install the latest version (5.0) but got this errors:

image

Because these specific projects are dependant on the 3.1.16 version, so I found it better if I only upgrade my package to the 3.1.16 version.

Please let me know if makes sense to you.

jovinson-ms: We should be able to use the `SdkPackageVersion` variable that's set in [Directory.Build.props](https://github.com/microsoft/healthcare-shared-components/blob/b40438aaffbfe0da20d037cfe1a83a11ede177e6/Directory.Build.props#L27) - it will determine the correct version to use based on the framework. If you look at the [.csproj file](https://github.com/microsoft/healthcare-shared-components/blob/b40438aaffbfe0da20d037cfe1a83a11ede177e6/src/Microsoft.Health.SqlServer/Microsoft.Health.SqlServer.csproj#L23) you'll see this pattern being used.

jovinson-ms:
I should have mentioned this in the issue - `SqlClient` is only used in the `SchemaManager` tool which does it's own dependency injection [here](https://github.com/microsoft/healthcare-shared-components/blob/b40438aaffbfe0da20d037cfe1a83a11ede177e6/tools/SchemaManager/Program.cs#L47) - we'll need to add a call to `services.AddHttpClient()` for the program to be able to run. This should be it!

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
wbalbo: Okay, I changed the .csproj in the Http reference to follow this pattern, thanks.

jovinson-ms (approved):

wbalbo:
Thank you for all the tips and for the patience, Josiah, I really appreciate it!

jovinson-ms:
> Thank you for all the tips and for the patience, Josiah, I really appreciate it!

You're welcome - thanks so much for your contribution! Going forward, I'll be identifying some small issues in this repo and in Microsoft\dicom-server - would be glad to have your contributions there.

wbalbo:
> > Thank you for all the tips and for the patience, Josiah, I really appreciate it!
> 
> You're welcome - thanks so much for your contribution! Going forward, I'll be identifying some small issues in this repo and in Microsoft\dicom-server - would be glad to have your contributions there.

And I'll be glad to help in the best way that I can!

PR #335, opened 2021-10-19
Added FhirAnonymousOperationAttribute to catch FhirOperation linked to anonymous calls

apurvabhaleMS:
## Description
Today, we log AuditEventType as FhirOperation in RequestMetrics. For anonymous endpoint calls like Metadata or Versions we log {"Authentication":"","Operation":"","ResourceType":""}. With this change we added  FhirAnonymousOperationAttribute which extends AllowAnonymousAttribute and can retrieve the Fhir operation. For metadata call, in request metric we should see an entry as {"Authentication":"","Operation":"metadata","ResourceType":""}.

## Related issues
Addresses [[74172](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74172)].

## Testing
Deploy personal environment to DF. Make anonymous calls. Check RequestMetric logs for the respective endpoint call,

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
+++ b/src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNetCore.Mvc.Controllers;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Health.Api.Extensions;
+using Microsoft.Health.Api.Features.AnonymousOperation;
 
 namespace Microsoft.Health.Api.Features.Audit
 {
@@ -59,6 +60,11 @@ namespace Microsoft.Health.Api.Features.Audit
                 return auditEventTypeAttribute.AuditEventType;
             }
 
+            if (attribute is FhirAnonymousOperationAttribute fhirAnonymousOperationAttribute)
+            {
+                return fhirAnonymousOperationAttribute.FhirOperation;
+            }
+
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -70,7 +76,8 @@ public Task StartAsync(CancellationToken cancellationToken)
                 .Select(ad =>
                 {
                     Attribute attribute = ad.MethodInfo?.GetCustomAttributes<AllowAnonymousAttribute>().FirstOrDefault() ??
```
feordin, line 78: As FhirAnonymoussOperationAttribute is derived from AnonymousAttribute, will the first condition in this chain return nonnull?

I am also curious why the AuditEventTypeAttribute has the (Attribute) cast, but it is unnecessary elsewhere?

src/Microsoft.Health.Core/Features/Context/IRequestContext.cs
```diff
@@ -22,6 +22,9 @@ public interface IRequestContext
 
         string RouteName { get; set; }
 
+        // We want to log anonymous calls (FhirAnonymousOperationType) in RequestMetric with valid operation type e.g. Metadata and Versions and not just limit to standard AuditEventTypes
+        // This property holds value for both the AuditEventTypes and FhirAnonymousOperationType
+        // We will still continue to log audit logs for AuditEventTypes only. This change is made in AuditHelper class
```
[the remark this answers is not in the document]
apurvabhaleMS: AuditHelper changes will be included in the Fhir server changes

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
apurvabhaleMS: yes, it does return nonnull value. or in our specific case it returns metadata/versions.
For (Attribute) cast, I am also not so sure butI think that's because of ?? operator - it requires both the sides to be of same type. When we have FhirAnonymousOperationAttribute (which is derived from AllowAnonymousAttribute) and AllowAnonymousAttribute -> it doesn't require any casting may be because one is derived form the other. But when we compare AuditEventTypeAttribute with remaining two it does require casting.

feordin (approved):

PR #342, opened 2021-11-02
Increase timeout on Sql serverConnection

rbans96:
## Description
This PR sets the explicit statement timeout on ServerConnection. The default timeout is 10 mins which is not enough for few long running statements in the migration scripts.

## Related issues
Addresses [#86103](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=86103)

## Testing
This is tested manually by upgrading the schema for 1TB database from schema version 17 to 19.

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
@@ -51,5 +51,10 @@ namespace Microsoft.Health.SqlServer.Configs
         /// If set, the client id of the managed identity to use when connecting to SQL, if AuthenticationType == ManagedIdentity.
         /// </summary>
         public string ManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Specifies the statement timeout to set on ServerConnection.
+        /// </summary>
+        public int StatementTimeoutInSecs { get; set; } = 14400;
     }
 }
--- a/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
+
 namespace Microsoft.Health.SqlServer.Configs
 {
     public class SqlServerDataStoreConfiguration
@@ -55,6 +57,6 @@ namespace Microsoft.Health.SqlServer.Configs
         /// <summary>
         /// Specifies the statement timeout to set on ServerConnection.
         /// </summary>
-        public int StatementTimeoutInSecs { get; set; } = 14400;
+        public TimeSpan StatementTimeout { get; set; } = TimeSpan.FromSeconds(14400);
     }
 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
```diff
@@ -51,5 +51,10 @@ public class SqlServerDataStoreConfiguration
         /// If set, the client id of the managed identity to use when connecting to SQL, if AuthenticationType == ManagedIdentity.
         /// </summary>
         public string ManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Specifies the statement timeout to set on ServerConnection.
+        /// </summary>
+        public int StatementTimeoutInSecs { get; set; } = 14400;
```
brendankowitz, line 58: I think a TimeSpan would be better, but check if that works with config as expected

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
```diff
@@ -45,7 +54,13 @@ public async Task ExecuteScriptAndCompleteSchemaVersionAsync(string script, int
                 }
 
                 var server = new Server(serverConnection);
+                var watch = Stopwatch.StartNew();
+                _logger.LogInformation($"Script execution started at {DateTime.UtcNow}");
```
brendankowitz, line 58: nit: `Clock.UtcNow`

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
```diff
@@ -146,5 +161,13 @@ public async Task<bool> InstanceSchemaRecordExistsAsync(CancellationToken cancel
             using var command = new SqlCommand(procedureQuery, connection);
             return (int)await command.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false) != 0;
         }
+
+        private ServerConnection GetServerConnectionWithTimeout(SqlConnection sqlConnection)
+        {
+            var serverConnection = new ServerConnection(sqlConnection);
+            serverConnection.StatementTimeout = _sqlServerDataStoreConfiguration.StatementTimeoutInSecs;
```
bbatjargal, line 168: Just for my understanding, Is serverConnection.StatementTimeout same as server.ConnectionContext.StatementTimeout (updating it before line 60)? For my test, I used "server.ConnectionContext.StatementTimeout".

src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
bbatjargal, line 58: Is this config accessible from the schema manager command-line tool and the schema manager Paas?

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
rbans96: Yes, statementTimeout is property on serverConnection only. When we set StatementTimeout to server.ConnectionContext, its also adding timeout to serverConnection itself.

src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
rbans96: This config is for oss command line tool. Once these changes gets merged and in a version upgrade PR for healthcare-shared-components nugets in health-paas, we will need to add this parameter [here](https://microsofthealth.visualstudio.com/Health/_git/health-paas?path=/src/SqlManagementApplication/SchemaManager.Paas/Commands/ApplyPaasCommand.cs&version=GBmaster&line=86&lineEnd=87&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

rbans96: Thanks, updated!

bbatjargal (approved):

PR #363, opened 2021-11-12
Publish symbols to Azure Artifacts

jovinson-ms:
## Description
Publish symbols to Azure Artifacts as part of CI.

## Related issues
Addresses [AB#78417](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78417).


build/package.yml
```diff
@@ -21,4 +21,11 @@ steps:
...
       pathtoPublish: '$(build.artifactStagingDirectory)/nupkgs'
       artifactName: 'nuget'
-      publishLocation: 'container'
\ No newline at end of file
+      publishLocation: 'container'
+
+  - task: PublishSymbols@2
+    displayName: 'Publish Symbols'
+    inputs:
+      searchPattern: '$(build.artifactStagingDirectory)\**\*.pdb'
```
wsugarman, line 29: `indexSources` needs to be `false`, right?

wsugarman, line 29: nit: can set `symbolsFolder: $(build.artifactStagingDirectory)` with `searchPattern: '**/*.pdb'`

wsugarman, line 29: What is the default `<DebugType>` in msbuild for the new SDK format? If it's not portable, then we should probably change to `<DebugType>portable</DebugType>`

jovinson-ms: I'm assuming it's portable because that's what gets built currently - will look for documentation.

jovinson-ms: `indexSources` defaults to true, and should be [per the documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/symbols?view=azure-devops#publish-symbols-to-azure-artifacts-symbol-server)

wsugarman: But I think you want [this doc](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/symbols?view=azure-devops#publish-portable-pdbs-to-azure-artifacts-symbol-server) as we're using SourceLink for indexing

jovinson-ms: yes - https://github.com/dotnet/sdk/blob/8c81f32aafb7b46678444f5354bc73dcde75a664/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L76

jovinson-ms: done

jovinson-ms: ok - updated the build task to match

build/build.yml
```diff
@@ -17,7 +17,7 @@ steps:
   displayName: 'dotnet build $(buildConfiguration)'
   inputs:
     command: build
-    arguments: '--configuration $(buildConfiguration) -p:AssemblyVersion="$(assemblySemVer)" -p:FileVersion="$(assemblySemFileVer)" -p:InformationalVersion="$(informationalVersion)" -p:ContinuousIntegrationBuild=true'
+    arguments: '--configuration $(buildConfiguration) -p:AssemblyVersion="$(assemblySemVer)" -p:FileVersion="$(assemblySemFileVer)" -p:InformationalVersion="$(informationalVersion)" -p:ContinuousIntegrationBuild=true -p:SourceLinkCreate=true'
```
wsugarman, line 20: I don't think `SourceLinkCreate` is actually needed

wsugarman (approved):

PR #364, opened 2021-11-12
Migrate to .NET 6.0

wsugarman:
## Description
Migrate `net5.0` projects to `net6.0`
- [x] Update *global.json*
- [x] Update target frameworks
  - [x] Shift multi-targeting from 3.1 + 5 to 5 + 6
- [x] Address new analyzer concerns
  - [x] Remove log messages from resource files
  - [x] Add exceptions to *.editorconfig*
- [x] Fix code generation to account for passing values by `ref`

## Related issues
[AB#86627](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86627)

## Testing
Tested locally and via pipelines

Breaking


tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.targets
```diff
@@ -2,22 +2,22 @@
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <PropertyGroup>
-        <GeneratorPath>$(MSBuildThisFileDirectory)../../tools/$(TargetFramework)/Microsoft.Health.Extensions.BuildTimeCodeGenerator.dll</GeneratorPath>
-    </PropertyGroup>
+  <PropertyGroup>
+    <GeneratorPath>$(MSBuildThisFileDirectory)../../tools/net6.0/Microsoft.Health.Extensions.BuildTimeCodeGenerator.dll</GeneratorPath>
```
wsugarman, line 6: We don't necessarily want the user's `TargetFramework` to change our tool's location. A user with the latest dotnet SDK can run `net6.0` applications via `dotnet <dll>`. Furthermore, once `net7.0` releases, this should still work even in `net7.0` projects.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/TypeExtensions.cs
```diff
@@ -27,6 +27,11 @@ public static TypeSyntax ToTypeSyntax(this Type t, bool useGlobalAlias = false)
                 return SyntaxFactory.PredefinedType(SyntaxFactory.Token(SyntaxKind.VoidKeyword));
             }
 
+            if (t.IsByRef)
+            {
+                return SyntaxFactory.RefType(t.GetElementType().ToTypeSyntax());
```
wsugarman, line 32: `ref` parameters

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/DelegatingInterfaceImplementationGenerator.cs
```diff
@@ -155,9 +155,13 @@ private IEnumerable<MemberDeclarationSyntax> GetPropertiesAndMethods()
                             SyntaxKind.SimpleMemberAccessExpression,
                             typedFieldName,
                             methodName),
-                        ArgumentList(SeparatedList(methodInfo.GetParameters().Select(p => Argument(IdentifierName(p.Name))))));
+                        ArgumentList(
```
wsugarman, line 158: `ref` arguments

brendankowitz:
I am excited to see work being done to move onto .net6, but a few questions on my mind. Do you have a plan on how/when we merge this? What is the compatibility with projects on .net5, do we all need to update together if we are no longer multitargeting?

For visibility: @adrianwyatt @cunninghamjc @dustinburson @abhijeett

wsugarman:
To answer your questions:
1. I am in the middle of updating the DICOM OSS project to `net6.0` in addition to updating these packages. In truth, I don't really need these changes to update to `net6.0` in the DICOM repo as these packages should be compatible. The only real change necessary is to update the SQL code gen tool's target file, as it incorrectly uses its consuming project's `$(TargetFramework)` for choosing the DLL to run (which there isn't a `net6.0` DLL to run)
2. `net5.0` projects cannot leverage `net6.0` packages, but *most* `net6.0` assemblies should be able to leverage `net5.0` packages. There are a number of [breaking changes](https://docs.microsoft.com/en-us/dotnet/core/compatibility/6.0) however that should be reviewed before using `net5.0` packages as a `net6.0` assembly.
3. We don't necessarily need to update together. Ultimately the questions become: (1) how quickly do other projects need changes from these packages beyond what's already been published and (2) how long would it take to update those libraries to `net6.0`. We can maintain the multi-target for `net5.0` and `net6.0` with some extra work, especially in the code gen as it appears `StringBuilder` has some new APIs. But if other projects do not need changes from these packages until they update to `net6.0` too, then keeping the multi-target for them would be unnecessary.

That being said, maintaining the `net6.0` target until `net8.0` may behoove us as it is the latest LTS version.

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaManagerDataStore.cs
```diff
@@ -56,12 +56,12 @@ public async Task ExecuteScriptAndCompleteSchemaVersionAsync(string script, int
 
                 var server = new Server(serverConnection);
                 var watch = Stopwatch.StartNew();
-                _logger.LogInformation($"Script execution started at {Clock.UtcNow}");
+                _logger.LogInformation("Script execution started at {UtcTime}", Clock.UtcNow);
```
brandonpollett, line 59: Nice catch

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -61,7 +61,7 @@ private async Task HandlerAsync(string connectionString, Uri server, MutuallyExc
 
         private bool EnsureForce()
         {
-            _logger.LogWarning(Resources.ForceWarning);
+            _logger.LogWarning("Are you sure to apply command with force option? Type 'yes' to confirm.");
```
brandonpollett, line 64: Interesting... I think we should probably track a backlog item to clean this up. We should really separate out what we want to be written to the console in an interactive manner vs what we want to log which should likely go to a file sink in this context.

brandonpollett (approved):
The code looks good to me once we resolve the discussion about readiness.

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Abstractions used by Microsoft Health.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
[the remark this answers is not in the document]
brendankowitz: Are there other projects that we could change to Standard too? It doesn't seem like a bad direction for Share Components

Directory.Build.props
```diff
@@ -9,11 +9,11 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <HighEntropyVA>true</HighEntropyVA>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>latest</LangVersion>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <Product>Microsoft Health</Product>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <SupportedFrameworks>netcoreapp3.1;net5.0</SupportedFrameworks>
```
brendankowitz, line 16: Just wondering, why not keep the variable and change to `net6.0`?

wsugarman: I've actually gone back and added it back. I've also figured out how to multi-target the code-gen stuff.

wsugarman:
Updated package:
image

src/Microsoft.Health.Abstractions/Microsoft.Health.Abstractions.csproj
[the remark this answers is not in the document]
wsugarman: Perhaps! But it can become difficult to use standard here if they use the microsoft extension libs. Let's say a lib wants to target standard and uses the latest 6.0.0 version of `Microsoft.Extensions.Logging`. However, it this lib is also used by our packages (or apps) that use ASP.NET Core, a .net 5 project would use the 5.0.0 of these packages. This makes it all a bit hairy.

src/Microsoft.Health.Extensions.DependencyInjection/Microsoft.Health.Extensions.DependencyInjection.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Dependency injection extensions for Microsoft's default implementation used by Microsoft Health.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
wsugarman, line 5: Will change back to `netstandard2.0`

tools/SchemaManager.Core/SchemaManager.Core.csproj
```diff
@@ -1,9 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(SupportedFrameworks);</TargetFrameworks>
+    <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
     <AssemblyName>Microsoft.Health.SchemaManager.Core</AssemblyName>
-    <OutputType>Library</OutputType>
```
wsugarman, line 6: This is the default

Directory.Build.props
```diff
@@ -9,11 +9,12 @@
...
+    <SdkPackageVersion>6.0.0</SdkPackageVersion>
+    <SupportedFrameworks>net5.0;net6.0</SupportedFrameworks>
```
Ivanidzo4ka, line 17: >>net5.0;net6.0 [](http://example.com/codeflow?start=24&length=14)

.net 3 and .net 6 are Long term support, while .net5 is not.
[https://dotnet.microsoft.com/platform/support/policy/dotnet-core#lifecycle](https://dotnet.microsoft.com/platform/support/policy/dotnet-core#lifecycle)

I would throw away .net 5 and leave support for .net 3 and .net 6.

wsugarman: If both FHIR and DICOM both don't have any libraries still running `netcoreapp3.1`, I'm inclined to remove it. But I think we need to keep `net5.0` in case migrations to `net6.0` take a bit. Is there any FHIR code that needs `netcoreapp3.1`?

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.targets
```diff
@@ -3,9 +3,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <ToolTargetFramework>net5.0</ToolTargetFramework>
-    <ToolTargetFramework Condition="'$(TargetFramework)' == 'netcoreapp3.1' Or '$(TargetFramework)' == 'net5.0'">$(TargetFramework)</ToolTargetFramework>
-    <GeneratorPath>$(MSBuildThisFileDirectory)../../tools/$(ToolTargetFramework)/Microsoft.Health.Extensions.BuildTimeCodeGenerator.dll</GeneratorPath>
+    <GeneratorPath>$(MSBuildThisFileDirectory)../../tools/net6.0/Microsoft.Health.Extensions.BuildTimeCodeGenerator.dll</GeneratorPath>
   </PropertyGroup>
```
Ivanidzo4ka, line 9: >net6 [](http://example.com/codeflow?start=58&length=4)

I think this would make generator to be unusable for anything other than .net6

wsugarman: It will still work with `net5.0`, as the `ToolTargetFramework` will update for either `net5.0` or `net6.0`. However, in the future if we wanted to use this tool with a `net7.0` assembly, the target file would choose the `net6.0` assembly to use.

Ivanidzo4ka: maybe I'm using generator in a wrong way...

Error MSB3073 The command "dotnet "C:\Users\ivmatan\.nuget\packages\microsoft.health.extensions.buildtimecodegenerator\3.2.38\build\netstandard2.0\../../tools/net6.0/Microsoft.Health.Extensions.BuildTimeCodeGenerator.dll" --generator-name "ImmutableSqlModelGenerator" --output-file "C:\src\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Schema\Model\ImmutableTypes.Generated.cs" --namespace "Microsoft.Health.Fhir.SqlServer.Features.Schema.Model" " exited with code 1.

Ivanidzo4ka (approved):
:shipit:

Directory.Build.props
Ivanidzo4ka: No, we don't and you right it's make sense to have one LTS version and previous one rather than two LTS

PR #388, opened 2022-01-13
Make DefaultAzureCredentialOptions configurable

jnlycklama:
## Description
DefaultAzureCredentialOptions added to the configurations - now for deploy we can specify which authentication flows should be used. So for testing, any auth can be used, but prod can specify to use a single authentication flow.

## Related issues
Addresses #87826.

## Testing
Tested that config would pick up changes made in dicom app settings

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
@@ -20,12 +20,16 @@ namespace Microsoft.Health.Blob.Configs
         /// <summary>
         /// If set, the client id of the managed identity to use when connecting to azure storage, if AuthenticationType == ManagedIdentity.
         /// </summary>
-        public string ManagedIdentityClientId { get; set; }
+        public string ManagedIdentityClientId
+        {
+            get => Credentials.ManagedIdentityClientId;
+            set => Credentials.ManagedIdentityClientId = value;
+        }
 
         /// <summary>
         /// Gets or sets the options for configuring DefaultAzureCredential
         /// </summary>
         /// <value>The settings for configuring the default azure credential</value>
-        public DefaultAzureCredentialOptions DefaultAzureCredentialOptions { get; set; } = new DefaultAzureCredentialOptions();
+        public DefaultAzureCredentialOptions Credentials { get; set; } = new DefaultAzureCredentialOptions();
     }
 }
--- a/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using Azure.Identity;
+
 namespace Microsoft.Health.Blob.Configs
 {
     public class BlobDataStoreConfiguration
@@ -19,5 +21,11 @@ namespace Microsoft.Health.Blob.Configs
         /// If set, the client id of the managed identity to use when connecting to azure storage, if AuthenticationType == ManagedIdentity.
         /// </summary>
         public string ManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the options for configuring DefaultAzureCredential
+        /// </summary>
+        /// <value>The settings for configuring the default azure credential</value>
+        public DefaultAzureCredentialOptions DefaultAzureCredentialOptions { get; set; } = new DefaultAzureCredentialOptions();
     }
 }
--- a/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
+++ b/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
@@ -26,7 +26,8 @@ namespace Microsoft.Health.Blob.Features.Storage
 
             if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
             {
-                ManagedIdentityCredential credential = new ManagedIdentityCredential(configuration.ManagedIdentityClientId);
+                configuration.DefaultAzureCredentialOptions.ManagedIdentityClientId = configuration.ManagedIdentityClientId;
+                DefaultAzureCredential credential = new DefaultAzureCredential(configuration.DefaultAzureCredentialOptions);
 
                 return new BlobServiceClient(new Uri(configuration.ConnectionString), credential, blobClientOptions);
             }
--- a/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
+++ b/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
@@ -26,8 +26,7 @@ namespace Microsoft.Health.Blob.Features.Storage
 
             if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
             {
-                configuration.DefaultAzureCredentialOptions.ManagedIdentityClientId = configuration.ManagedIdentityClientId;
-                DefaultAzureCredential credential = new DefaultAzureCredential(configuration.DefaultAzureCredentialOptions);
+                DefaultAzureCredential credential = new DefaultAzureCredential(configuration.Credentials);
 
                 return new BlobServiceClient(new Uri(configuration.ConnectionString), credential, blobClientOptions);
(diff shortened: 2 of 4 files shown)


src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
```diff
@@ -19,5 +21,11 @@ public class BlobDataStoreConfiguration
         /// If set, the client id of the managed identity to use when connecting to azure storage, if AuthenticationType == ManagedIdentity.
         /// </summary>
         public string ManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the options for configuring DefaultAzureCredential
+        /// </summary>
+        /// <value>The settings for configuring the default azure credential</value>
+        public DefaultAzureCredentialOptions DefaultAzureCredentialOptions { get; set; } = new DefaultAzureCredentialOptions();
```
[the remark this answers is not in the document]
wsugarman: We could also refactor `ManagedIdentityClientId ` to serve as a proxy for `DefaultAzureCredentialOptions.ManagedIdentityClientId`.
Eg.
```csharp
public string ManagedIdentityClientId
{
    get => DefaultAzureCredentialOptions.ManagedIdentityClientId;
    set => DefaultAzureCredentialOptions.ManagedIdentityClientId = value;
}
```

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -131,9 +131,11 @@ public static IServiceCollection AddBlobServiceClient(this IServiceCollection se
                     IAzureClientBuilder<BlobServiceClient, BlobClientOptions> clientBuilder;
                     if (string.Equals(options.Credential, "managedidentity", StringComparison.OrdinalIgnoreCase))
```
wsugarman, line 132: If this is all configuration based, perhaps this can be simplified to always specify the `WithCredential` since we're using the `DefaultCredentials`?

wsugarman (changes requested):

src/Microsoft.Health.Blob/Configs/BlobDataStoreConfiguration.cs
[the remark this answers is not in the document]
wsugarman: Or maybe since we may always use this, `Credentials` is a good name?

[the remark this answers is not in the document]
jnlycklama: Yes, but with the bools flipped. Like this - https://microsofthealth.visualstudio.com/Health/_git/workspace-platform/pullrequest/17213?_a=files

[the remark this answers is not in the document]
jnlycklama: I like the proxy option! Updated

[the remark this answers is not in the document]
jnlycklama: I like just "Credentials", updated

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -131,9 +131,11 @@ public static IServiceCollection AddBlobServiceClient(this IServiceCollection se
...
                     {
+                        options.DefaultAzureCredentialOptions.ManagedIdentityClientId = options.ClientId;
```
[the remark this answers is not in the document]
jnlycklama: Updated to use the proxy

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
jnlycklama: There's code above to set `options.ConnectionString = BlobLocalEmulator.ConnectionString;` for testing purposes, and that won't be able to transform into a Uri below. In theory we could get rid of the Uri transformation and just always pass the connectionstring into `AddBlobServiceClient` as a string, but I'm not sure if there are any other implications

src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
```diff
@@ -26,7 +26,7 @@ public static BlobServiceClient Create(BlobDataStoreConfiguration configuration)
 
             if (configuration.AuthenticationType == BlobDataStoreAuthenticationType.ManagedIdentity)
             {
-                DefaultAzureCredential credential = new DefaultAzureCredential(new DefaultAzureCredentialOptions { ManagedIdentityClientId = configuration.ManagedIdentityClientId });
+                DefaultAzureCredential credential = new DefaultAzureCredential(configuration.Credentials);
```
wsugarman, line 29: Should we always use these credentials, regardless of managed identity?

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -133,7 +133,7 @@ public static IServiceCollection AddBlobServiceClient(this IServiceCollection se
                     {
                         clientBuilder = builder
                             .AddBlobServiceClient(new Uri(options.ConnectionString))
-                            .WithCredential(new DefaultAzureCredential(new DefaultAzureCredentialOptions { ManagedIdentityClientId = options.ClientId }));
+                            .WithCredential(new DefaultAzureCredential(options.Credentials));
```
wsugarman, line 136: Like in my other comment, should we always use the credentials and instead just check for managed identity to determine which overload of `AddBlobServiceClient` to call?

wsugarman: But you're right that we do need to use the different overloads. It's very non-obvious though lol. The string overload uses the key-value pair connection strings like `DefaultEndpointsProtocol=https;AccountName=storagesample;AccountKey=<account-key>`, while `Uri` overload takes the base URL for all azure storage account services.

wsugarman (changes requested):

src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
jnlycklama: Looks like BlobServiceClient assumes that a string connectionstring will contain its own credentials. It doesn't give the option to use string + credentials in the constructor https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobserviceclient.-ctor?view=azure-dotnet#azure-storage-blobs-blobserviceclient-ctor(system-string-azure-storage-blobs-blobclientoptions)

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
jnlycklama: Updated!

src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs
wsugarman: Oh you know, that makes perfect sense.

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
```diff
@@ -132,15 +132,16 @@ public static IServiceCollection AddBlobServiceClient(this IServiceCollection se
...
-                            .WithCredential(new DefaultAzureCredential(new DefaultAzureCredentialOptions { ManagedIdentityClientId = options.ClientId }));
+                            .AddBlobServiceClient(new Uri(options.ConnectionString));
                     }
                     else
                     {
                         clientBuilder = builder
                             .AddBlobServiceClient(options.ConnectionString);
                     }
 
+                    clientBuilder = clientBuilder.WithCredential(new DefaultAzureCredential(options.Credentials));
```
wsugarman, line 143: Given your other comment, I think you were right originally to keep this only in the managed identity scenario, as the credentials are included as part of the `string` otherwise.

wsugarman (approved):
I think the PR looks good, but we should probably only use the `DefaultAzureCredential` object when a non-connection string is used. Today that's only for managed identity

src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs
jnlycklama: Makes sense! Reverted my last change

PR #397, opened 2022-01-26
Fix sql connection transient errors

smithago:
## Description

Fix transient issues seen in production issue by setting the ConnectRetryCount.

- Rename ISqlConnectionFactory to ISqlConnectionBuilder, this is not a factory but has 2 actual type implementations.
- Move common SqlConnection logic to helper class
- Set ConnectRetryCount from default 1 to default 3 and ConnectRetryTimeOut = 30sec

More info on it below:

https://docs.microsoft.com/en-us/azure/azure-sql/database/troubleshoot-common-connectivity-issues
https://docs.microsoft.com/en-us/sql/connect/ado-net/step-4-connect-resiliently-sql-ado-net?view=sql-server-ver15

## Related issues
[AB#88154](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88154).

## Testing
Existing test pass


src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (string.IsNullOrEmpty(sqlConnectionString))
+            {
+                throw new InvalidOperationException("The SQL connection string cannot be null or empty.");
+            }
+
+            var connectionStringBuilder = new SqlConnectionStringBuilder(sqlConnectionString);
+
+            // https://docs.microsoft.com/en-us/azure/azure-sql/database/troubleshoot-common-connectivity-issues
+            // Change from default 1 to 3
+            connectionStringBuilder.ConnectRetryCount = sqlServerTransientFaultRetryPolicyConfiguration.ConnectRetryCount;
```
brandonpollett, line 45: I think if we do this, we need to also increase the timeout per https://docs.microsoft.com/en-us/azure/azure-sql/database/troubleshoot-common-connectivity-issues#net-sqlconnection-parameters-for-connection-retry

src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+            string sqlConnectionString = await sqlConnectionStringProvider.GetSqlConnectionString(cancellationToken);
+            if (string.IsNullOrEmpty(sqlConnectionString))
+            {
+                throw new InvalidOperationException("The SQL connection string cannot be null or empty.");
+            }
+
+            var connectionStringBuilder = new SqlConnectionStringBuilder(sqlConnectionString);
+
+            // https://docs.microsoft.com/en-us/azure/azure-sql/database/troubleshoot-common-connectivity-issues
+            // Change from default 1 to 3
```
Ivanidzo4ka, line 44: >3 [](http://example.com/codeflow?start=40&length=1)

supernit:
I think your default value is 4.
```ConnectRetryCount { get; set; } = 4```

File: src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs [hunk 3]
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Health.SqlServer.Configs;
+
+namespace Microsoft.Health.SqlServer
+{
+    /// <summary>
+    /// Helper class to build the base SqlConnection object
```
Ivanidzo4ka, line 15: >SqlConnection [](http://example.com/codeflow?start=39&length=13)

supernit:
you can use <see cref="SqlConnection"/> and VS and any other editor would pick it up in case of renaming the class.

Otherwise it's just a string, and you need to update string manually in case of refactoring.

src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs
Ivanidzo4ka: Is it worth to expose all three properties (ConnectRetryCount, ConnectRetryInterval, ConnectionTimeout) via config with some reasonable defaults?

I don't see reason to include only one of them and not rest.

src/Microsoft.Health.SqlServer/ISqlConnection.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Health.SqlServer
 {
-    public interface ISqlConnectionFactory
+    public interface ISqlConnection
```
Ivanidzo4ka, line 12: >ISqlConnection [](http://example.com/codeflow?start=21&length=14)

maybe offtopic, but I think it was a decent name, and changing to just connection made things less readable.

```using SqlConnection connection = await _sqlConnection.GetSqlConnectionAsync()```
vs 
```using SqlConnection connection = await _sqlConnectionFactory.GetSqlConnectionAsync```

first one is less characters, but it's confusing, why I need to call `_sqlConnection `to create `SqlConnection connection` object.

Comment (smithago) [reply src/Microsoft.Health.SqlServer/ISqlConnection.cs:12]: Good point, but a factory.Get() returns an interface that can have multiple implementation.

DI is doing the Factory work. 

This class is AuthenticationType Specific SqlConnection builder. I can name this to ISqlConnectionBuilder and name the other class I introduced as SqlConnectionHelper or something.  Any suggestions?

File: src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs (hunk 1 above)
Comment (smithago) [reply src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs]: Fixed, thanks for catching

File: src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs (hunk 2 above)
Comment (smithago) [reply src/Microsoft.Health.SqlServer/SqlConnectionBuilder.cs]: fixed

File: src/Microsoft.Health.SqlServer/ISqlConnection.cs (hunk 4 above)
Comment (smithago) [reply src/Microsoft.Health.SqlServer/ISqlConnection.cs]: Take a look now, let me know

Comment (Ivanidzo4ka) [reply src/Microsoft.Health.SqlServer/ISqlConnection.cs]: Honestly as long as there are meaningful word after connection, it should be good. Otherwise it blurs between connection itself and object which provide connection.

`ConnectionBuilder` sounds lovely to me.

**Review by Ivanidzo4ka — APPROVED** [review]
:shipit:

File: src/Microsoft.Health.SqlServer/Configs/SqlServerTransientFaultRetryPolicyConfiguration.cs [hunk 5]
```diff
@@ -18,7 +18,7 @@ public class SqlServerTransientFaultRetryPolicyConfiguration
         public TimeSpan InitialDelay { get; set; } = TimeSpan.FromMilliseconds(100);
 
         /// <summary>
-        /// The maximum number of retries to use, in addition to the original call.
+        /// The maximum number of retries to use, in addition to the original call for SqlCommand Execute*.
```
Ivanidzo4ka, line 21: >* [](http://example.com/codeflow?start=105&length=1)

supernit: this `*` looks strange here. Probably typo.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -69,14 +69,25 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
 
             services.AddOptions();
             services.TryAddSingleton<ISqlConnectionStringProvider, DefaultSqlConnectionStringProvider>();
-            services.TryAddSingleton<ISqlConnectionFactory>(
+
+            services.TryAddSingleton<ISqlConnectionBuilder>(
```
Ivanidzo4ka, line 73: May I ask to add comment why we try to add two `ISqlConnectionBuilders` here, and how they different?

smithago: sure

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -69,14 +69,27 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
...
+                     SqlServerDataStoreConfiguration config = sqlServerDataStoreConfigOption.Value;
+                     ISqlConnectionStringProvider sqlConnectionStringProvider = p.GetRequiredService<ISqlConnectionStringProvider>();
+                     return config.AuthenticationType == SqlServerAuthenticationType.ManagedIdentity
+                         ? new ManagedIdentitySqlConnectionBuilder(sqlConnectionStringProvider, p.GetRequiredService<IAccessTokenHandler>(), sqlServerDataStoreConfigOption)
+                         : new DefaultSqlConnectionBuilder(sqlConnectionStringProvider, sqlServerDataStoreConfigOption);
+                 });
+
+            // We build SqlConnection object differently for different Authentication Type.
+            // For ManagedIdentity access we set the access token
+            services.TryAddSingleton<ISqlConnectionBuilder>(
```
brendankowitz, line 86: Wouldn't the second`services.TryAddSingleton<ISqlConnectionBuilder>` do nothing if it is already registered above?

PR #416, opened 2022-02-25
Refactor Health Check Cache to Support New Features

wsugarman:
## Description
Refactored the `CachedHealthCheck` to support:
- Skipping failed results in cache such that they can be retried more quickly
- A "refresh offset" that enables health check results to be considered stale, and such that they may be refreshed without invalidating them quite yet
- If a result is still valid, it can be returned without contention over the semaphore

Also added support for setting the timeout across all health checks

## Related issues
N/A

## Testing
New tests

Feature (although technically breaking)


src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -10,91 +10,133 @@
...
+        private volatile CachedHealthCheckResult _cachedResult = new CachedHealthCheckResult();
 
-            _provider = provider;
-            _healthCheck = healthCheck;
-            _logger = logger;
+        public CachedHealthCheck(IServiceProvider provider, Func<IServiceProvider, IHealthCheck> healthCheck)
+        {
+            _provider = EnsureArg.IsNotNull(provider, nameof(provider));
+            _healthCheck = EnsureArg.IsNotNull(healthCheck, nameof(healthCheck));
+            _options = provider.GetRequiredService<IOptions<HealthCheckCachingOptions>>()?.Value;
```
brandonpollett, line 33: Does this mean that all health checks types are cached using the same options for a service?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -10,91 +10,133 @@
...
+                    return _cachedResult.Value;
+                }
             }
-            catch (OperationCanceledException oce) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException oce) when (oce.CancellationToken == cancellationToken)
             {
-                _logger.LogInformation(oce, $"Cancellation was requested for {nameof(CheckHealthAsync)}");
-                return _lastResult;
+                _logger.LogWarning(oce, $"Cancellation was requested for {nameof(CheckHealthAsync)}.");
+                throw;
```
brandonpollett, line 66: I guess it doesn't matter because the operation is cancelled, but what response does this actually return?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -10,91 +10,133 @@
...
-                await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                // Note: Only the timespan controls whether or not the method returns true/false.
+                //       If the token is canceled, the method throws instead an OperationCanceledException.
+                if (!await _semaphore.WaitAsync(maxWait, cancellationToken).ConfigureAwait(false))
+                {
+                    return _cachedResult.Value;
+                }
             }
-            catch (OperationCanceledException oce) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException oce) when (oce.CancellationToken == cancellationToken)
```
brandonpollett, line 63: What is the semantic difference between checking if the cancellationToken is the same as we were passed versus checking IsCancellationRequested? Also, what happens if this isn't the case/can that happen?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -10,91 +10,133 @@
...
 
-            _provider = provider;
-            _healthCheck = healthCheck;
-            _logger = logger;
+        public CachedHealthCheck(IServiceProvider provider, Func<IServiceProvider, IHealthCheck> healthCheck)
+        {
+            _provider = EnsureArg.IsNotNull(provider, nameof(provider));
+            _healthCheck = EnsureArg.IsNotNull(healthCheck, nameof(healthCheck));
+            _options = provider.GetRequiredService<IOptions<HealthCheckCachingOptions>>()?.Value;
+            _logger = provider.GetRequiredService<ILogger<CachedHealthCheck>>();
```
brandonpollett, line 34: Just a thought, but we should try to restrict the use of this to the `PostConfigure`. I think it makes more sense to have it there since it's more of a common pattern there.

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
brandonpollett: Would it make sense to return the existing status if we're going to throw it away anyway?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -10,91 +10,133 @@
...
+            return _cachedResult;
+        }
+
+        private bool IsCacheFresh(DateTimeOffset currentTime)
+            => currentTime < _cachedResult.StaleTime && (_options.CacheFailure || _cachedResult.Value.Status != HealthStatus.Unhealthy);
+
         public void Dispose()
         {
-            Dispose(true);
+            _semaphore.Dispose();
```
brandonpollett, line 129: Is there a chance for the `_semaphore` to be null? (Just asking because previously that was checked)

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckTimeoutPostConfigure.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.Health.Api.Features.HealthChecks
+{
+    internal sealed class HealthCheckTimeoutPostConfigure : IPostConfigureOptions<HealthCheckServiceOptions>
+    {
+        private readonly TimeSpan _timeout;
+
+        public HealthCheckTimeoutPostConfigure(TimeSpan timeout)
+            => _timeout = EnsureArg.IsGte(timeout, TimeSpan.Zero, nameof(timeout));
+
+        public void PostConfigure(string name, HealthCheckServiceOptions options)
```
brandonpollett, line 20: Is there any order to the execution of the two post configure options?

brandonpollett: Or I guess more so, is there any guarantee that they'll be registered in the correct order?

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -29,6 +29,8 @@ public SqlServerHealthCheck(SqlConnectionWrapperFactory sqlConnectionWrapperFact
 
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken)
         {
+            _logger.LogInformation($"Starting {nameof(SqlServerHealthCheck)}.");
```
brandonpollett, line 32: 🎉

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Yeah, they all share the same options. We could in the future add settings per health check (using a dictionary keyed by the name of the health check). But I don't think it's necessary right now

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: It returns "Unhealthy". In the case of a user's cancellation, they'll never see the response since they aborted the connection. But if we timeout (and the result has expired), we'll throw. Perhaps I can change the catch block later to also check for a fresh result.

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Eh, probably not an important one lol. In fact, we should probably not check this at all!

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Perhaps? We may want to keep an options object in the health check

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Nope. It gets initialized before the ctor

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckTimeoutPostConfigure.cs
wsugarman: I am unsure if there is a way to guarantee order... I'll check! At this point they are presumably based on the order in which they are added to the service container

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Updated

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Removed

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: Move the DI to the post configure

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckTimeoutPostConfigure.cs
wsugarman: Removed timeout from the cache specifically

src/Microsoft.Health.Api.UnitTests/Features/HealthCheck/HealthCheckTimeoutPostConfigureTests.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.Extensions.Options;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Api.UnitTests.Features.HealthCheck
+{
+    public class HealthCheckTimeoutPostConfigureTests
+    {
+        [Fact]
+        public void GivenHealthChecks_WhenOverridingTimeout_ThenReplaceTimeout()
```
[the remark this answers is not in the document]
wsugarman: Removed

brandonpollett (approved):

PR #420, opened 2022-03-04
Allow Multiple Refresh Threads in Health Check Cache

wsugarman:
## Description
- Add new `MaxRefreshThreads` which controls how many threads can concurrently refresh
- Move validation from multiple places into a single class that is added in the module
- Refactor cache creation

## Related issues
N/A

## Testing
Locally

Patch


src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -35,52 +35,64 @@ internal sealed class CachedHealthCheck : IHealthCheck, IDisposable
             _healthCheckFactory = EnsureArg.IsNotNull(healthCheckFactory, nameof(healthCheckFactory));
             _options = EnsureArg.IsNotNull(options, nameof(options));
             _logger = EnsureArg.IsNotNull(logger, nameof(logger));
+            _semaphore = new SemaphoreSlim(_options.MaxRefreshThreads, _options.MaxRefreshThreads);
```
smithago, line 38: Call looks odd to me. Passing max for both init and max.
Should we just use constructor with initialCount single param?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -35,52 +35,64 @@ internal sealed class CachedHealthCheck : IHealthCheck, IDisposable
...
         }
 
         public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
-            DateTimeOffset currentTime = Clock.UtcNow;
-            if (IsCacheFresh(currentTime))
+            CacheSnapshot current = GetCacheSnapshot();
```
smithago, line 43: Sorry if I am missing some pre-context here. But I was expecting something like

return GetHealthCheckResult();

Basically the understanding of caching and refreshing is done in GetHealthCheckResult()
{
   var result = _cachedResult;

   If (result == null || result.IsExpired)
   {
       RefreshCache()
       result = _cachedResult[]
   }
  return result.Result;
}

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -94,58 +106,78 @@ private async Task<HealthCheckResult> RefreshCache(HealthCheckContext context, D
...
+            // The cache is only updated if appropriate
+            if (current.CanUpdate(newResult.Status, _options.CacheFailure))
             {
-                _cachedResult = new CachedHealthCheckResult
+                var newCache = new CachedHealthCheckResult
                 {
-                    ExpireTime = currentTime + _options.Expiry,
-                    StaleTime = currentTime + _options.Expiry - _options.RefreshOffset,
-                    Value = result,
+                    ExpireTime = current.Timestamp + _options.Expiry,
```
smithago, line 140: What is the difference between Expired vs Stale? Do we need this complication?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -35,52 +35,64 @@ internal sealed class CachedHealthCheck : IHealthCheck, IDisposable
...
+            CacheSnapshot current,
+            TimeSpan maxSemaphoreWait,
+            CancellationToken cancellationToken)
+        {
             try
             {
                 // Note: Only the timespan controls whether or not the method returns true/false.
                 //       If the token is canceled, the method throws instead an OperationCanceledException.
-                if (!await _semaphore.WaitAsync(maxWait, cancellationToken).ConfigureAwait(false))
+                if (!await _semaphore.WaitAsync(maxSemaphoreWait, cancellationToken).ConfigureAwait(false))
```
smithago, line 71: There are so many try catch here, it is hard to follow. Can this be composed to something like

GetHealthCheckFromOrigin() // Get actual sql health check.

RefreshCache()
{
    semaphore.Wait();
    try
    {
        GetHealthCheckFromOrigin();
        UpdateCache();
    }
    finally
    {
        semaphore.Release();
    }
}

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckCachingOptions.cs
```diff
@@ -17,19 +16,22 @@ public class HealthCheckCachingOptions
...
         /// </summary>
-        [Range(typeof(TimeSpan), "00:00:00", "1.00:00:00", ConvertValueInInvariantCulture = true, ParseLimitsInInvariantCulture = true)]
         public TimeSpan Expiry { get; set; } = TimeSpan.FromSeconds(1);
 
         /// <summary>
         /// Gets or sets the amount of time before the expiry for which the instance of <see cref="HealthCheckResult"/>
         /// should be refreshed.
         /// </summary>
-        [Range(typeof(TimeSpan), "00:00:00", "1.00:00:00", ConvertValueInInvariantCulture = true, ParseLimitsInInvariantCulture = true)]
         public TimeSpan RefreshOffset { get; set; } = TimeSpan.Zero;
```
smithago, line 25: I am not sure how this is used?

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckCachingOptions.cs
```diff
@@ -17,19 +16,22 @@ public class HealthCheckCachingOptions
...
 
         /// <summary>
         /// Gets or sets a flag that indicates whether <see cref="HealthStatus.Unhealthy"/> should be cached.
         /// </summary>
         public bool CacheFailure { get; set; } = true;
```
smithago, line 30: Do we need to allow this option? Should'nt we always cache failure? We don't want to retry a failed sql ops immediately either.

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckCachingOptions.cs
```diff
@@ -17,19 +16,22 @@ public class HealthCheckCachingOptions
...
+
+        /// <summary>
+        /// Gets or sets the maximum number of threads that can refresh the cache concurrently.
+        /// </summary>
+        public int MaxRefreshThreads { get; set; } = 1;
```
smithago, line 35: Was this going to be 2?

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: It isn't necessary, but it does add a level of correctness. The only way for us to extend beyond the initial count would for an extra `Release()` invocation be made, but that could not happen without `Enter()`.

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: I attempted to simply it a bit more

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: As discussed offline, it may be helpful to attempt to refresh before the cache fully expires

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
wsugarman: This wait is a bit tricky. The code path for exiting via timeout is when the cache hasn't expired. Threads will attempt to enter the semaphore and give up quickly if someone else is already refreshing. Exiting via exception means that the caller has abandoned waiting on the health check entirely

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckCachingOptions.cs
wsugarman: Discussed offline

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckCachingOptions.cs
wsugarman: Removed

src/Microsoft.Health.Api/Features/HealthChecks/HealthCheckCachingOptions.cs
wsugarman: Done

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -7,148 +7,139 @@
...
+                    _logger.LogError(ex, "Health check failed to complete. Falling back to cache.");
                 }
 
-                return _cachedResult.Value;
+                return _cache.Result;
             }
 
-            return UpdateCache(result).Value;
+            // Update cache based on the latest snapshot
+            return RefreshCache(result);
```
brandonpollett, line 98: It may make it more clear to have the `RefreshCache` just do it's updating the `_cache` object and just return `_cache.Result` like we do elsewhere?

brandonpollett (dismissed):

brandonpollett (approved):

PR #421, opened 2022-03-05
Base the retry on the new SqlClient RetryLogicProvider

smithago:
## Description
New SqlClient after 4.00 version provides inbuilt retry for transient sql issues
https://docs.microsoft.com/en-us/sql/connect/ado-net/configurable-retry-logic-sqlclient-introduction?view=sql-server-ver15
https://devblogs.microsoft.com/azure-sql/configurable-retry-logic-for-microsoft-data-sqlclient/

## Related issues
[AB#88206](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88206)

## Testing
Existing tests pass, new unit tests added


src/Microsoft.Health.SqlServer/Features/Client/RetrySqlCommandWrapperFactory.cs
```diff
@@ -14,21 +13,20 @@ namespace Microsoft.Health.SqlServer.Features.Client
...
-            _retryPolicy = sqlTransientFaultRetryPolicyFactory.Create();
+            EnsureArg.IsNotNull(sqlRetryLogicBaseProvider, nameof(sqlRetryLogicBaseProvider));
+            _sqlRetryLogicBaseProvider = sqlRetryLogicBaseProvider;
         }
 
         /// <inheritdoc/>
         public override SqlCommandWrapper Create(SqlCommand sqlCommand)
         {
             return new RetrySqlCommandWrapper(
                 base.Create(sqlCommand),
```
wsugarman, line 28: In the future, we'd like users to invoke `SqlConnection.OpenAsync` before they execute their commands, but do we need to include this inside of the `Execute*Async*` methods for `RetrySqlCommandWrapper` now that we can rely on the framework to handle re-opening the connection?

If not, we can probably remove the logic to open the connection explicitly again in the `RetrySqlCommandWrapperFactory` and perhaps even just delete `RetrySqlCommandWraper`. We can instead change this factory to do something like:

```csharp
sqlCommand.RetryLogicProvider = _sqlRetryLogicBaseProvider;
return new SqlCommandWrapper(base.Create(sqlCommand));
```

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -48,9 +49,27 @@ public static class SqlServerBaseRegistrationExtensions
...
+
+                    // Preferred gap time to delay before retry
+                    DeltaTime = TimeSpan.FromSeconds(config.TransientFaultRetryPolicy.ConnectRetryIntervalInSeconds),
+
+                    // Maximum gap time for each delay time before retry
+                    MaxTimeInterval = TimeSpan.FromSeconds(config.TransientFaultRetryPolicy.ConnectMaxTimeIntervalInSeconds),
+                };
+
+                // Create a retry logic provider
+                return SqlConfigurableRetryFactory.CreateExponentialRetryProvider(options);
```
wsugarman, line 71: We could also introduce a new configuration for these values to avoid having to map these values ourselves. Something like:

```csharp
public class SqlClientRetryOptions
{
    public SqlRetryStrategy Strategy { get; set; } // Maps to the various factory methods like None, Exponential, Incremental, and Fixed
    public SqlRetryLogicOption Settings { get; set; } // Unfortunately this is a sealed class! Otherwise, I would say let's extend it
}
```

Then we could allow users to configure these options themselves based on hard-coded values or a config via:
```csharp
services.Configure(x =>
{
    x.NumberOfTries = 12;
});

// OR

services
    .AddOptions<SqlClientRetryOptions>()
    .Bind(config.GetSection("SqlServer:Retry"));
```

With this class, we can update the retry factory to generate a `SqlRetryLogicBaseProvider` in its ctor based on these options:

```csharp
public RetrySqlCommandWrapperFactory(IOptions<SqlClientRetryOptions> options)
{
    SqlClientRetryOptions o = EnsureArg.IsNotNull(options?.Value, nameof(options));
    _retryProvider = o.Strategy switch
    {
        SqlRetryStrategy.None => SqlConfigurableRetryFactory.CreateNoneRetryProvider(),
        SqlRetryStrategy.Fixed => SqlConfigurableRetryFactory.CreateFixedRetryProvider(o.Settings),
        SqlRetryStrategy.Incremental => SqlConfigurableRetryFactory.CreateIncrementalRetryProvider(o.Settings),
        SqlRetryStrategy.Exponential => SqlConfigurableRetryFactory.CreateExponentialRetryProvider(o.Settings),
        _ => throw new ArgumentOutOfRangeException(nameof(options)),
    };
}
```

smithago: Thanks Will. I like the pattern of reusing the same objects. I also want to be more prescreptive in the framework. All the clients of the SDK should inherit a default retry with an option to override. Mostly to help pass best practices across all clients.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -187,5 +173,35 @@ public static IServiceCollection AddSqlServerManagement<TVersion>(this IServiceC
...
+        private static IServiceCollection AddSqlRetryLogicProvider(this IServiceCollection services)
+        {
+            services.TryAddSingleton<SqlRetryLogicBaseProvider>(p =>
+            {
+                SqlServerDataStoreConfiguration config = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value;
+
+                // set default
+                if (config.Retry == null)
+                {
+                    config.Retry = new SqlClientRetryOptions();
```
wsugarman, line 192: We can move this to `SqlServerDataStoreConfiguration.cs`, right? By default, the property can be populated

src/Microsoft.Health.SqlServer/SqlConnectionHelper.cs
```diff
@@ -40,22 +39,18 @@ internal class SqlConnectionHelper
...
-            // Change from default 10 to 10
-            connectionStringBuilder.ConnectRetryInterval = sqlServerTransientFaultRetryPolicyConfiguration.ConnectRetryIntervalInSeconds;
-
             if (initialCatalog != null)
             {
                 connectionStringBuilder.InitialCatalog = initialCatalog;
             }
 
-            return new SqlConnection(connectionStringBuilder.ToString());
+            var sqlConnection = new SqlConnection(connectionStringBuilder.ToString());
```
wsugarman, line 47: nit:
```suggestion
            return new SqlConnection(connectionStringBuilder.ToString()) { RetryLogicProvider = sqlRetryLogic };
```

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
wsugarman: And optionally if someone really wants it to be `null` (because they explicitly set it in their config), then we could treat it like `None`

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -84,13 +83,34 @@ internal async Task InitializeAsync(CancellationToken cancellationToken, bool op
             }
         }
 
+        [Obsolete("Please use " + nameof(CreateRetrySqlCommand) + " or " + nameof(CreateNonRetrySqlCommand) + " instead.")]
         public SqlCommandWrapper CreateSqlCommand()
         {
             SqlCommand sqlCommand = SqlConnection.CreateCommand();
+            sqlCommand.Transaction = SqlTransaction;
```
wsugarman, line 90: nit: just have this return `CreateRetrySqlCommand()`

src/Microsoft.Health.SqlServer/Features/Client/RetrySqlCommandWrapper.cs
```diff
@@ -8,74 +8,56 @@
...
         /// <inheritdoc/>
-        public override Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
+        public override async Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
         {
-            return _retryPolicy.ExecuteAsync(async () =>
-            {
-                await EnsureConnectionOpenAsync(cancellationToken);
-                return await _sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
-            });
+           await EnsureConnectionOpenAsync(cancellationToken);
```
wsugarman, line 30: Do we still need this here?

wsugarman: If we don't, then we could probably also delete this class and simply assign the `SqlRetryLogicBaseProvider` inside of `SqlConnectionWrapper` when generating a retryable command

src/Microsoft.Health.SqlServer/Features/Client/SqlCommandWrapper.cs
```diff
@@ -95,6 +88,15 @@ public SqlTransaction Transaction
             set => _sqlCommand.Transaction = value;
         }
 
+        /// <summary>
+        /// <see cref="SqlCommand.RetryLogicProvider"/>.
+        /// </summary>
+        public SqlRetryLogicBaseProvider RetryLogicProvider
```
wsugarman, line 94: I think this is now unnecessary, right?

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -187,5 +173,29 @@ public static IServiceCollection AddSqlServerManagement<TVersion>(this IServiceC
...
+        /// <param name="services">The <see cref="IServiceCollection"/> to be updated.</param>
+        /// <returns>The <paramref name="services"/> for additional method invocations.</returns>
+        /// <exception cref="NotImplementedException">When the retry mode is unkown</exception>
+        private static IServiceCollection AddSqlRetryLogicProvider(this IServiceCollection services)
+        {
+            services.TryAddSingleton<SqlRetryLogicBaseProvider>(p =>
+            {
+                SqlServerDataStoreConfiguration config = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value;
+
+                return config.Retry.Mode switch
```
wsugarman, line 189: nit: Should we null-check? Like `config.Retry?.Mode ?? SqlRetryMode.None`

wsugarman (approved):

src/Microsoft.Health.SqlServer/Features/Client/SqlCommandWrapper.cs
smithago: yes, thanks for catching

wsugarman (approved):

PR #428, opened 2022-03-14
Use File-Scoped Namespaces

wsugarman:
## Description
Bulk update every C# file to use file-scoped namespaces and enforce it via .editorconfig

## Related issues
N/A

## Testing
Pipeline

+semver: skip


.editorconfig
```diff
@@ -43,6 +43,9 @@ csharp_style_var_for_built_in_types = false:none
 csharp_style_var_when_type_is_apparent = true:suggestion
 csharp_style_var_elsewhere = false:suggestion
 
+# use file-scoped namespaces
+csharp_style_namespace_declarations = file_scoped:warning
```
wsugarman, line 47: This is the change that actually enforces the style

PR #429, opened 2022-03-14
Refactor Build

wsugarman:
## Description
- Remove stylecop and migrate rules to *.editorconfig*
- Refactor build pipelines
- Use latest 1ES tasks

## Related issues
N/A

## Testing
Pipelines

+semver: skip


src/Microsoft.Health.Api/Features/Audit/AuditException.cs
```diff
@@ -3,14 +3,15 @@
...
+using System.Globalization;
 using Microsoft.Health.Abstractions.Exceptions;
 
 namespace Microsoft.Health.Api.Features.Audit;
 
 public class AuditException : MicrosoftHealthException
 {
     public AuditException(string controllerName, string actionName)
-        : base(string.Format(Resources.MissingAuditInformation, controllerName, actionName))
+        : base(string.Format(CultureInfo.CurrentCulture, Resources.MissingAuditInformation, controllerName, actionName))
```
wsugarman, line 14: Generally any client-facing messages, like those localized via the resource files, should be formatted via `CultureInfo.CurrentCulture`. For example, you can see this pattern in the BCLs via the `SR` class which encapsulates "System Resources." See [Format](https://github.com/dotnet/runtime/blob/4ec7a81108d944f292c7886653fc8f42494adc9d/src/libraries/Common/src/System/SR.cs#L55)

It is called without an `IFormatProvider`, which is the same as "current culture" according to the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=net-6.0#format-and-culture)

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -31,22 +31,14 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
-
-            await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false);
-
-            _logger.LogInformation("Successfully connected to SQL database.");
-            return HealthCheckResult.Healthy("Successfully connected.");
-        }
-        catch (Exception ex)
-        {
-            _logger.LogError(ex, "Failed to connect to the data store.");
-            return HealthCheckResult.Unhealthy("Failed to connect.");
```
wsugarman, line 49: We don't need to catch this. An uncaught exception will return 503 with a generic "Unhealthy" status message

src/Microsoft.Health.SqlServer/Features/Schema/Messages/Get/GetCompatibilityVersionRequest.cs
```diff
@@ -5,8 +5,12 @@
 
 using MediatR;
 
+#pragma warning disable CA1716 // Identifiers should not match keywords
+
 namespace Microsoft.Health.SqlServer.Features.Schema.Messages.Get;
```
wsugarman, line 10: The problem lies with `Get` as the namespace

src/Microsoft.Health.SqlServer/Features/Schema/Model/ColumnUtilities.cs
```diff
@@ -21,7 +22,7 @@ internal static long GetLengthForFloatColumn(byte precision)
         }
         else
         {
-            throw new ArgumentOutOfRangeException(string.Format("Precision {0} must be between 1 & 53", precision));
+            throw new ArgumentOutOfRangeException(string.Format(CultureInfo.CurrentCulture, "Precision {0} must be between 1 & 53", precision));
```
brendankowitz, line 25: Why `CurrentCulture` in some places and `InvariantCulture` in others?

src/Microsoft.Health.SqlServer/Features/Schema/Messages/Get/GetCompatibilityVersionResponse.cs
```diff
@@ -19,3 +21,5 @@ public GetCompatibilityVersionResponse(CompatibleVersions versions)
 
     public CompatibleVersions CompatibleVersions { get; }
 }
+
+#pragma warning restore CA1716 // Identifiers should not match keywords
```
brendankowitz, line 25: Is this needed? i.e. is it just file scoped anyway?

src/Microsoft.Health.SqlServer/Features/Schema/Model/ColumnUtilities.cs
wsugarman: I've updated the cultures such that `CurrentCulture` is used for client-facing messages, like exception messages, while `InvariantCulture` is used for our own log messages and parsing.

src/Microsoft.Health.SqlServer/Features/Schema/Messages/Get/GetCompatibilityVersionResponse.cs
wsugarman: Oh, I suppose that's true!

brendankowitz (approved):
LGTM

PR #438, opened 2022-03-22
Add Common Function Packages

wsugarman:
## Description
- Add three new libraries:
  1. `Microsoft.Health.Operations` - Defines long-running operation abstractions that are agnostic of the platform
  2. `Microsoft.Health.Operations.Functions` - Includes Azure Function-specific types related to the execution of long-running operations; also includes a few commonly used Azure Functions
  3. `Microsoft.Health.Functions.Extensions` - Utilities related speficially to Azure Functions and unrelated to long-running operations
- Add new tests, including Integration tests
  - Add Azurite to pipeline tests for azure function integration tests

## Related issues
Addresses [AB#87713](https://microsofthealth.visualstudio.com/Health/_workitems/edit/87713)

## Testing
Tested via new unit and integration tests

Feature


build/integration-tests.yml
```diff
@@ -1,6 +1,9 @@
 # We create a separate template for these tests because the hosted Windows agents do not support Linux docker containers,
 # and the MS SQL Server team does not provide a docker image whose base image is Windows. Therefore these steps can only run on Linux agents.
 steps:
+- script: docker run -p 10000:10000 -p 10001:10001 -p 10002:10002 -d --rm --name azure-storage mcr.microsoft.com/azure-storage/azurite:latest
```
wsugarman, line 4: We now run functions integration tests with azurite as the backend. We pass metadata to the tests via environment variables which are loaded as configuration key/values pairs

src/Microsoft.Health.Functions.Extensions/AzureFunctionsJobHostBuilder.cs
```diff
@@ -0,0 +1,133 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.Functions.Extensions.Configuration;
+
+namespace Microsoft.Health.Functions.Extensions;
+
+/// <summary>
+/// Represents a builder for instances of <see cref="IHost"/> that run Azure Functions.
+/// </summary>
+public sealed class AzureFunctionsJobHostBuilder
```
wsugarman, line 24: This type is meant to be used when running integration tests. Check out the `WebJobsTestFixture` for an E2E example of its usage

test/Microsoft.Health.Functions.Examples/host.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "2.0",
+    "Extensions": {
+        "DurableTask": {
+            "HubName": "IntegrationTests"
+        }
+    },
+    "Logging": {
+        "FileLoggingMode": "never",
+        "LogLevel": {
```
wsugarman, line 10: I updated the log levels to try not to spam the test output

test/Microsoft.Health.Functions.Examples/Sorting/DistributedSorter.cs
```diff
@@ -0,0 +1,93 @@
...
+
+public class DistributedSorter
+{
+    private readonly SortingOptions _options;
+
+    public DistributedSorter(IOptions<SortingOptions> options)
+        => _options = EnsureArg.IsNotNull(options?.Value, nameof(options));
+
+    [FunctionName(nameof(InsertionSortAsync))]
+    public async Task<IReadOnlyList<int>> InsertionSortAsync([OrchestrationTrigger] IDurableOrchestrationContext context, ILogger logger)
```
wsugarman, line 26: This is a pretty basic orchestration, but the hope is that it illustrates an durable function that calls multiple activities and even restarts via `ContinueAsNew`

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Microsoft.Health.Operations.Functions.Management;
+
+/// <summary>
+/// Represents a periodic process that purges stale orchestration instance metadata from storage.
+/// </summary>
+public sealed class PurgeOrchestrationInstanceHistory
+{
+    private readonly PurgeHistoryOptions _options;
+
+    private const string PurgeFrequencyVariable = $"%{AzureFunctionsJobHost.RootSectionName}:{PurgeHistoryOptions.SectionName}:{nameof(PurgeHistoryOptions.Frequency)}%";
```
wsugarman, line 25: Unfortunately this means that the configuration path is hard-coded, so all uses must use this path in their configs.

src/Microsoft.Health.Operations/Serialization/OperationIdJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Microsoft.Health.Operations.Serialization;
+
+/// <summary>
+/// Represents a strict <see cref="JsonConverter{T}"/> for operation IDs.
+/// </summary>
+/// <remarks>
+/// Unlike the typical <see cref="Guid"/> deserialization, the <see cref="OperationIdJsonConverter"/>
+/// only accepts strings that can be parsed using the <see cref="OperationId.FormatSpecifier"/>.
+/// </remarks>
+public sealed class OperationIdJsonConverter : JsonConverter<Guid>
```
wsugarman, line 20: As a note, there are two serialization libraries in play: json.net and system.text.json.

For all types internal to our activities and orchestrations, we rely on json.net (aka newtonsoft) as that is the library used within the Azure Functions framework. However, our code (like the serialization between the client and server for our web apps), will use the newer System.Text.Json.

src/Microsoft.Health.Operations/Microsoft.Health.Operations.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
wsugarman, line 1: We could also call this `Microsoft.Health.Operations.Abstractions`

test/Microsoft.Health.Functions.Examples/Sorting/DistributedSorter.cs
```diff
@@ -0,0 +1,93 @@
...
+        else
+        {
+            logger.LogInformation("Sorting complete: [{Values}]", string.Join(", ", checkpoint.Values));
+        }
+
+        return checkpoint.Values;
+    }
+
+    [FunctionName(nameof(SortRange))]
+    public Task<int[]> SortRange([ActivityTrigger] int[] values, ILogger logger)
```
shivakumar-ms, line 70: could we consider changing the parameter and return types to IEnumerable<int>?

test/Microsoft.Health.Functions.Examples/Sorting/DistributedSorter.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        for (int j = values.Length - 1; j > 0 && comparer.Compare(values[j - 1], values[j]) > 0; j--)
+        {
+            (values[j - 1], values[j]) = (values[j], values[j - 1]);
+        }
+
+        return Task.FromResult(values);
+    }
+
+    private static T[] Concat<T>(T[] left, T[] right)
```
shivakumar-ms, line 85: could we consider changing the parameter and return types to IEnumerable<T>?

test/Microsoft.Health.Functions.Examples/Sorting/DistributedSorter.cs
wsugarman: If it's `IEnumerable<T>`, then I cannot perform the swap in this method, and I would need to re-allocate the elements as an array to mutate it ☹️

test/Microsoft.Health.Functions.Examples/Sorting/DistributedSorter.cs
wsugarman: I am unsure changing between `T[]` and `IEnumerable<T>` is helpful when ultimately I need to mutate the structure. Keeping the input as arrays allows for the more performant `Array.Copy` method

src/Microsoft.Health.Operations/Microsoft.Health.Operations.csproj
wsugarman: I tried to rename, but I didn't like it when needing the abstractions namespace for `OperationId`

test/Microsoft.Health.Functions.Examples/host.json
```diff
@@ -0,0 +1,29 @@
...
+            "default": "Information",
+            "DurableTask": "Warning",
+            "Function": "Warning",
+            "Function.*.User": "Information",
```
wsugarman, line 14: I had no idea this works lol

pengchen0692 (dismissed):

src/Microsoft.Health.Operations/OperationState.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Gets or sets the operation ID.
+    /// </summary>
+    /// <value>The unique ID that denotes a particular operation.</value>
+    public Guid OperationId { get; init; }
+
+    /// <summary>
+    /// Gets or sets the category of the operation.
+    /// </summary>
+    public T? Type { get; init; }
```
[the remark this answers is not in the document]
wsugarman: We have an `enum` for DICOM, but I imagine you could use `string`

pengchen0692 (approved):

PR #442, opened 2022-03-24
Revise the command timeout to 180 seconds

v-mserdar:
## Description
We're creating a temporary build for a customer that need to have the default CommandTimeout increased to 180 seconds.

## Related issues
n/a

## Testing
Successfully ran the following tests.
Microsoft.Health.SqlServer.Tests.E2E (net5.0)
﻿Microsoft.Health.SqlServer.Tests.E2E (net6.0)
﻿Microsoft.Health.SqlServer.UnitTests (net5.0)
﻿Microsoft.Health.SqlServer.UnitTests (net6.0)

Refs [AB#89901](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/89901)
+semver: patch

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Client/SqlCommandWrapper.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Client/SqlCommandWrapper.cs
@@ -26,6 +26,7 @@ public class SqlCommandWrapper : IDisposable
         EnsureArg.IsNotNull(sqlCommand, nameof(sqlCommand));
 
         _sqlCommand = sqlCommand;
+        CommandTimeout = 180;
     }
 
     /// <summary>


SergeyGaluzo (approved):
:shipit:

brendankowitz:
@smithago, @wsugarman any concerns with having this as a higher hardcoded/global limit for now?

Ideally I think the default should be configurable per app and overridable per request

brendankowitz:
@v-mserdar I'd say semver should be at least patch for this since its a slightly different behavior

v-mserdar:
> @v-mserdar I'd say semver should be at least patch for this since its a slightly different behavior

semver is updated to patch as suggested

feordin (approved):

PR #464, opened 2022-04-26
Throw truncate error if string value too long

feordin:
## Description
Compare string length to the max length of the column when setting the value.  Throw truncate error if the value is too long.

## Related issues
Addresses [issue [AB#91346](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91346)].

## Testing
Unit test was added.

Patch


wsugarman:
I was playing around with the connections for the PR pipeline, so if they fail to trigger correctly, please re-create the PR. Sorry about this!

feordin:
Will do 🙂.
________________________________

wsugarman (approved):

PR #485, opened 2022-05-12
AB#91764 - ISchemaClient Changes

mitchell-fox:
This PR removes the URI requirements from ISchemaClient and places them in the implementation of SchemaClient to remove the dependency on a running service for SchemaManager upgrades. 

The following changes have been made:
- ISchemaClient / SchemaClient
    - Removed SetUri(Uri) from interface (still implemented in SchemaClient though).
    - Changed GetScriptAsync(Uri) and GetDiffScriptAsync(Uri) to GetScriptAsync(int version) and GetDiffScriptAsync(int version), respectively.
- KnownRoutes
    - Split Script path into IdSegment and shorter Script path
    - Added RootedScriptUri and RootedDiffUri, which replace the IdSegment with a version number for SqlSchemaManager.
- Updated all tests in SqlSchemaManagerTests to new SqlSchemaManager definitions.
- ISchemaManager / SchemaManager
    - Removed "Uri server" parameter from all methods.
    - Updated GetScriptAsync() to use new "bool applyFullVersion" parameter to differentiate between upgrades from version 1 to another version in one big upgrade, or between versions 2+ with incremental upgrades.
- ApplyCommand / AvailableCommand / CurrentCommand
    - All commands now take a SchemaClient in their constructors and call SetUri(server).
        - Previously, SqlSchemaManager was in charge of calling SetUri(server).

All tests in healthcare-shared-components pass with these changes. However, there are only unit tests for SqlSchemaManager, and there are no integration tests that I could find.

I am working on testing the SchemaManager EXE manually but would like to open this PR for code review and any other suggestions/additions.


src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -27,7 +27,7 @@ public SchemaClient(IHttpClientFactory httpClientFactory)
 
     public void SetUri(Uri uri)
     {
-        if (_httpClient.BaseAddress == null)
+        if(_httpClient.BaseAddress == null)
```
wsugarman, line 30: nit: missing space

src/Microsoft.Health.SqlServer/Routing/KnownRoutes.cs
```diff
@@ -15,10 +16,17 @@ public static class KnownRoutes
...
-    public const string Script = Versions + "/{id:int}/script";
+    public const string IdSegment = "{id:int}";
+    public const string Script = Versions + "/" + IdSegment + "/script";
     public const string Diff = Script + "/diff";
 
     internal static readonly Uri RootedCurrentUri = new Uri("/" + SchemaRoot + "/" + Current, UriKind.Relative);
     internal static readonly Uri RootedCompatibilityUri = new Uri("/" + SchemaRoot + "/" + Compatibility, UriKind.Relative);
     internal static readonly Uri RootedVersionsUri = new Uri("/" + SchemaRoot + "/" + Versions, UriKind.Relative);
+
+    internal static Uri RootedScriptUri(int version) =>
```
wsugarman, line 27: nit: I think our style has been to put `=>` on the next line for methods

wsugarman, line 27: nit: This is a method and as such should start with a verb like `GetRootedScriptUri`. Same comment as below

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -27,7 +27,7 @@ public SchemaClient(IHttpClientFactory httpClientFactory)
...
         {
             _httpClient.BaseAddress = uri;
```
wsugarman, line 32: Why is the `BaseAddress` bound so late? Shouldn't we set the `BaseAddress` via DI such that the correct value is prsent by the time it is populated in the ctor? I know you didn't write this initially, do you know if this is possible?

src/Microsoft.Health.SqlServer/Routing/KnownRoutes.cs
```diff
@@ -15,10 +16,17 @@ public static class KnownRoutes
     public const string Versions = "versions";
 
     public const string Current = Versions + "/current";
-    public const string Script = Versions + "/{id:int}/script";
+    public const string IdSegment = "{id:int}";
```
wsugarman, line 19: Do you know why this is a route template? Is this because the server component is using this same class for its route? I'd probably prefer to have separate members for the client and server instead of trying to replace it to create a format string for the client

tools/SchemaManager.Core/ISchemaManager.cs
```diff
@@ -18,28 +18,25 @@ public interface ISchemaManager
     /// Applies SQL schemas specified by the range in <paramref name="type"/>.
     /// </summary>
     /// <param name="connectionString">connection string to the backing SQL server.</param>
-    /// <param name="server">URL to the service.</param>
     /// <param name="type">The schema version to apply.</param>
     /// <param name="token">A cancellation token.</param>
     /// <returns>A task.</returns>
-    public Task ApplySchema(string connectionString, Uri server, MutuallyExclusiveType type, CancellationToken token = default);
```
wsugarman, line 25: Do you know why `connectionString` is part of each method's signature and not part of the state of the schema manager? I'm not aware of schema manager communicating with different SQL servers

tools/SchemaManager.Core/SqlSchemaManager.cs
```diff
@@ -131,13 +128,15 @@ public virtual async Task ApplySchema(string connectionString, Uri server, Mutua
 
             if (availableVersions.First().Id == 1)
```
wsugarman, line 129: We should not use `First()` or `Last()` here. We have a list, so we know we have indices to perform random access. We can even use the newer syntax `[^1]` to access the last element.

tools/SchemaManager.Core/SqlSchemaManager.cs
```diff
@@ -273,14 +266,14 @@ private async Task ValidateVersionCompatibility(int maxAvailableVersion, Cancell
         }
     }
 
-    private async Task<string> GetScriptAsync(int version, string scriptUri, CancellationToken cancellationToken, string diffUri = null)
+    private async Task<string> GetScriptAsync(bool applyFullVersion, int version, CancellationToken cancellationToken)
```
wsugarman, line 269: Is this method necessary? Could we not simply call the `_schemaClient` methods directly?

tools/SchemaManager/Commands/ApplyCommand.cs
```diff
@@ -56,7 +60,9 @@ private async Task HandlerAsync(string connectionString, Uri server, MutuallyExc
             return;
         }
 
-        await _schemaManager.ApplySchema(connectionString, server, exclusiveType, cancellationToken).ConfigureAwait(false);
+        _schemaClient.SetUri(server);
+
+        await _schemaManager.ApplySchema(connectionString, exclusiveType, cancellationToken).ConfigureAwait(false);
```
wsugarman, line 65: nit: remove `await` and `async` from this method. We can simply return the `Task` itself

tools/SchemaManager/Commands/CurrentCommand.cs
```diff
@@ -39,6 +41,7 @@ public CurrentCommand(ISchemaManager schemaManager)
         EnsureArg.IsNotNull(schemaManager, nameof(schemaManager));
 
         _schemaManager = schemaManager;
+        _schemaClient = schemaClient;
```
wsugarman, line 44: Requires `null`-checking

tools/SchemaManager/Commands/AvailableCommand.cs
```diff
@@ -36,11 +38,14 @@ public AvailableCommand(ISchemaManager schemaManager)
         EnsureArg.IsNotNull(schemaManager);
 
         _schemaManager = schemaManager;
+        _schemaClient = schemaClient;
```
wsugarman, line 41: Requires `null`-checking

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
mitchell-fox: Thanks for noticing this; fixed it in my latest commit.

src/Microsoft.Health.SqlServer/Routing/KnownRoutes.cs
mitchell-fox: Fixed in latest commit.

mitchell-fox: Fixed in latest commit.

tools/SchemaManager/Commands/CurrentCommand.cs
mitchell-fox: Fixed in latest commit.

tools/SchemaManager/Commands/ApplyCommand.cs
mitchell-fox: Should I make this change in AvailableCommand.cs and CurrentCommand.cs as well?

tools/SchemaManager.Core/SqlSchemaManager.cs
mitchell-fox: Good catch! Definitely not necessary. Fixed in latest commit.

src/Microsoft.Health.SqlServer/Routing/KnownRoutes.cs
mitchell-fox: I'm guessing it's a route template because of its use in the AvailableVersion SQL Server API: https://github.com/microsoft/healthcare-shared-components/blob/eeb2beb077ba0798ae81e28745345aa23731fe72/src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs#L59

Can you please elaborate on your suggestion? I agree the format string approach isn't good, but I don't know of a better solution.

tools/SchemaManager.Core/ISchemaManager.cs
mitchell-fox: Addressed in latest commit(s).

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
mitchell-fox: Addressed in latest commit(s).

mitchell-fox:
I am currently adding some integration tests since SchemaManager will need it with these new changes.

tools/SchemaManager/Program.cs
```diff
@@ -43,18 +47,35 @@ private static Parser BuildParser(ServiceProvider serviceProvider)
...
+        var builder = new ConfigurationBuilder();
+
+        builder.AddCommandLine(args);
+
+        IConfigurationRoot config = builder.Build();
+
+        services.AddOptions<CommandLineOptions>().Configure(x =>
+        {
+            x.ConnectionString = config["connection-string"];
+            x.Server = new Uri(config["server"]);
```
brandonpollett, line 67: This line will error out if the user doesn't pass anything to the exe. Previously when you don't pass anything, you get the following output:
```
Required command was not provided.

Usage:
  Microsoft.Health.SchemaManager [options] [command]

Options:
  --version    Display version information

Commands:
  apply        Applies the specified SQL schema update to the supplied connection string for the given server.
  available    Returns the available SQL schema versions for the given server.
  current      Returns the current SQL schema version(s) from the given server.
```

brandonpollett: Also, not sure if this will work with the short versions of the parameters as seen below
```
You must include a connection string.

apply:
  Applies the specified SQL schema update to the supplied connection string for the given server.

Usage:
  Microsoft.Health.SchemaManager apply [options]

Options:
  -cs, --connection-string <CONNECTION-STRING>    The connection string of the SQL server to apply the schema update.
  -s, --server <SERVER>                           The root URL of the server to connect to.
  -v, --version <VERSION>                         The versions between current version and specified version to apply.
  -n, --next <NEXT>                               The next version to apply.
  -l, --latest <LATEST>                           The versions between current version and latest available version to apply.
  -f, --force <FORCE>                             The schema migration is run without validating the schema verison.
```

tools/SchemaManager/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "SchemaManager": {
       "commandName": "Project",
-      "commandLineArgs": "Microsoft.Health.SchemaManager apply --connection-string \"server=(local);Initial Catalog=SHARED_HEALTHCARE;Integrated Security=true\" --server https://localhost:63637 --version 3"
+      "commandLineArgs": "Microsoft.Health.SchemaManager apply --connection-string \"server=(local);Initial Catalog=Dicom;Integrated Security=true;TrustServerCertificate=True\" --server https://localhost:63838 --version 3"
```
brandonpollett, line 5: This will need to be reverted before checkin.

tools/SchemaManager/Program.cs
mitchell-fox: Thanks for noticing these issues. They have been fixed in the latest commit.

tools/SchemaManager/Properties/launchSettings.json
mitchell-fox: Thanks for noticing this. Corrected in latest commit.

mitchell-fox:
We are going to forego adding integration tests and rely on manual testing for now. In the future we will open a PR to add integration tests.

tools/SchemaManager/CommandCollectionExtensions.cs
```diff
@@ -44,4 +45,42 @@ public static IServiceCollection AddCliCommands(this IServiceCollection services
...
+
+            if (!string.IsNullOrWhiteSpace(config[OptionAliases.Server]))
+            {
+                x.Server = new Uri(config[OptionAliases.Server]);
+            }
+        });
+
+        return services;
+    }
+
```
brandonpollett, line 85: nit: extra line

brandonpollett (dismissed):

src/Microsoft.Health.SqlServer/Routing/KnownRoutes.cs
```diff
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Globalization;
```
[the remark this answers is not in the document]
mitchell-fox: Fixed in latest commits.

tools/SchemaManager/CommandCollectionExtensions.cs
mitchell-fox: Fixed in latest commits.

rbans96:
@mitchell-fox  Thanks for the changes. Looks good to me. You might already be aware but just calling out that we would need to refactor this package a little [here](https://microsofthealth.visualstudio.com/Health/_git/health-paas?path=/src/SqlManagementApplication/SchemaManager.Paas) to consume this change in paas.

brandonpollett (approved):

PR #489, opened 2022-05-12
Add support to skip purging instances

bcarthic:
## Description
This PR adds support to skip purging instances. This is useful if we want to keep the state of the orchestration in the table storage.

## Related issues
Addresses [[AB#91768](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91768)].

## Testing
Add unit tests

Patch

### DIFF
--- a/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
+++ b/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
@@ -8,7 +8,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using DurableTask.Core;
 using EnsureThat;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
@@ -82,7 +81,8 @@ public class PurgeOrchestrationInstanceHistoryTests
         _purgeConfig.InstancesToSkipPurging = new string[] { instanceId1 };
         _purgeTask = new PurgeOrchestrationInstanceHistory(Options.Create(_purgeConfig));
 
-        var durableOrchestrationState = new List<DurableOrchestrationStatus> {
+        var durableOrchestrationState = new List<DurableOrchestrationStatus> 
+        {
             new DurableOrchestrationStatus { InstanceId = instanceId1, Name = instanceId1 },
             new DurableOrchestrationStatus { InstanceId = instanceId2, Name = instanceId2 }
         };
--- a/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
+++ b/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
@@ -5,8 +5,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
+using EnsureThat;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
 using Microsoft.Extensions.Logging.Abstractions;
@@ -24,8 +27,8 @@ public class PurgeOrchestrationInstanceHistoryTests
     private readonly DateTime _utcNow;
     private readonly TimerInfo _timer;
     private readonly PurgeHistoryOptions _purgeConfig;
-    private readonly PurgeOrchestrationInstanceHistory _purgeTask;
     private readonly IDurableOrchestrationClient _durableClient;
+    private PurgeOrchestrationInstanceHistory _purgeTask;
 
     public PurgeOrchestrationInstanceHistoryTests()
     {
@@ -33,7 +36,7 @@ public class PurgeOrchestrationInstanceHistoryTests
         _timer = Substitute.For<TimerInfo>(default, default, default);
         _purgeConfig = new PurgeHistoryOptions
         {
-            Statuses = new HashSet<OrchestrationStatus> { OrchestrationStatus.Completed },
+            Statuses = new HashSet<OrchestrationRuntimeStatus> { OrchestrationRuntimeStatus.Completed },
             MinimumAgeDays = 14,
         };
         _purgeTask = new PurgeOrchestrationInstanceHistory(Options.Create(_purgeConfig));
@@ -43,14 +46,57 @@ public class PurgeOrchestrationInstanceHistoryTests
     [Theory]
     [InlineData(0)]
     [InlineData(12)]
-    public async Task GivenNoOrchestrationInstances_WhenPurgeCompletedDurableFunctionsHistory_ThenNoOrchestrationsPurgedAsync(int deleted)
+    public async Task GivenOrchestrationInstances_WhenPurgeCompletedDurableFunctionsHistory_ThenOrchestrationsPurgedAsync(int count)
     {
+        var instanceId = Guid.NewGuid().ToString();
+
+        var durableOrchestrationState = Enumerable.Repeat(new DurableOrchestrationStatus { InstanceId = instanceId }, count);
+
         _durableClient
-            .PurgeInstanceHistoryAsync(
-                DateTime.MinValue,
-                _utcNow.AddDays(-_purgeConfig.MinimumAgeDays),
-                _purgeConfig.Statuses)
-            .Returns(new PurgeHistoryResult(deleted));
+            .ListInstancesAsync(Arg.Any<OrchestrationStatusQueryCondition>(), Arg.Any<CancellationToken>())
+            .Returns(new OrchestrationStatusQueryResult
+            {
+                DurableOrchestrationState = durableOrchestrationState
+            });
+
+        _durableClient
+            .PurgeInstanceHistoryAsync(instanceId)
+            .Returns(new PurgeHistoryResult(count));
+
+        using (Mock.Property(() => ClockResolver.UtcNowFunc, () => _utcNow))
+        {
+            await _purgeTask.Run(_timer, _durableClient, NullLogger.Instance);
+        }
+
+        await _durableClient
+            .Received(count)
+            .PurgeInstanceHistoryAsync(instanceId);
+    }
+
+    [Fact]
+    public async Task GivenOrchestrationInstancesAndInstancesToSkipPurging_WhenPurgeCompletedDurableFunctionsHistory_ThenNoOrchestrationsPurgedAsync()
+    {
+        var instanceId1 = Guid.NewGuid().ToString();
+        var instanceId2 = Guid.NewGuid().ToString();
+
+        _purgeConfig.InstancesToSkipPurging = new string[] { instanceId1 };
+        _purgeTask = new PurgeOrchestrationInstanceHistory(Options.Create(_purgeConfig));
+
+        var durableOrchestrationState = new List<DurableOrchestrationStatus> {
+            new DurableOrchestrationStatus { InstanceId = instanceId1 },
+            new DurableOrchestrationStatus { InstanceId = instanceId2 }
+        };
+
+        _durableClient
+            .ListInstancesAsync(Arg.Any<OrchestrationStatusQueryCondition>(), Arg.Any<CancellationToken>())
+            .Returns(new OrchestrationStatusQueryResult
+            {
+                DurableOrchestrationState = durableOrchestrationState
+            });
+
+        _durableClient
+            .PurgeInstanceHistoryAsync(Arg.Any<string>())
+            .Returns(new PurgeHistoryResult(1));
 
         using (Mock.Property(() => ClockResolver.UtcNowFunc, () => _utcNow))
         {
@@ -59,9 +105,6 @@ public class PurgeOrchestrationInstanceHistoryTests
 
         await _durableClient
             .Received(1)
-            .PurgeInstanceHistoryAsync(
-                DateTime.MinValue,
-                _utcNow.AddDays(-_purgeConfig.MinimumAgeDays),
-                _purgeConfig.Statuses);
+            .PurgeInstanceHistoryAsync(Arg.Any<string>());
     }
 }
--- a/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
+++ b/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
@@ -76,6 +76,13 @@ public class PurgeOrchestrationInstanceHistoryTests
     [Fact]
     public async Task GivenOrchestrationInstancesAndInstancesToSkipPurging_WhenPurgeCompletedDurableFunctionsHistory_ThenNoOrchestrationsPurgedAsync()
     {
+        OrchestrationStatusQueryCondition condition = new OrchestrationStatusQueryCondition
+        {
+            CreatedTimeFrom = DateTime.MinValue,
+            CreatedTimeTo = _utcNow.AddDays(-_purgeConfig.MinimumAgeDays),
+            RuntimeStatus = _purgeConfig.Statuses,
+        };
+
         var instanceId1 = Guid.NewGuid().ToString();
         var instanceId2 = Guid.NewGuid().ToString();
 
@@ -83,19 +90,19 @@ public class PurgeOrchestrationInstanceHistoryTests
         _purgeTask = new PurgeOrchestrationInstanceHistory(Options.Create(_purgeConfig));
 
         var durableOrchestrationState = new List<DurableOrchestrationStatus> {
-            new DurableOrchestrationStatus { InstanceId = instanceId1 },
-            new DurableOrchestrationStatus { InstanceId = instanceId2 }
+            new DurableOrchestrationStatus { InstanceId = instanceId1, Name = instanceId1 },
+            new DurableOrchestrationStatus { InstanceId = instanceId2, Name = instanceId2 }
         };
 
         _durableClient
-            .ListInstancesAsync(Arg.Any<OrchestrationStatusQueryCondition>(), Arg.Any<CancellationToken>())
+            .ListInstancesAsync(condition, Arg.Any<CancellationToken>())
             .Returns(new OrchestrationStatusQueryResult
             {
                 DurableOrchestrationState = durableOrchestrationState
             });
 
         _durableClient
-            .PurgeInstanceHistoryAsync(Arg.Any<string>())
+            .PurgeInstanceHistoryAsync(instanceId2)
             .Returns(new PurgeHistoryResult(1));
 
         using (Mock.Property(() => ClockResolver.UtcNowFunc, () => _utcNow))
@@ -105,6 +112,6 @@ public class PurgeOrchestrationInstanceHistoryTests
 
         await _durableClient
             .Received(1)
-            .PurgeInstanceHistoryAsync(Arg.Any<string>());
+            .PurgeInstanceHistoryAsync(instanceId2);
     }
 }
--- a/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
+++ b/src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
@@ -48,18 +48,12 @@ public class PurgeOrchestrationInstanceHistoryTests
     [InlineData(12)]
     public async Task GivenOrchestrationInstances_WhenPurgeCompletedDurableFunctionsHistory_ThenOrchestrationsPurgedAsync(int count)
     {
-        OrchestrationStatusQueryCondition condition = new OrchestrationStatusQueryCondition
-        {
-            CreatedTimeFrom = DateTime.MinValue,
-            CreatedTimeTo = _utcNow.AddDays(-_purgeConfig.MinimumAgeDays),
-            RuntimeStatus = _purgeConfig.Statuses,
-        };
         var instanceId = Guid.NewGuid().ToString();
 
         var durableOrchestrationState = Enumerable.Repeat(new DurableOrchestrationStatus { InstanceId = instanceId }, count);
 
         _durableClient
-            .ListInstancesAsync(condition, Arg.Any<CancellationToken>())
+            .ListInstancesAsync(Arg.Is<OrchestrationStatusQueryCondition>(condition => AreConditionEqual(condition)), Arg.Any<CancellationToken>())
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Operations.Functions/Management/PurgeHistoryOptions.cs
```diff
@@ -43,4 +44,10 @@ public class PurgeHistoryOptions
     /// <value>A value cron expression</value>
     [Required]
     public string? Frequency { get; set; }
+
+    /// <summary>
+    ///  Gets or sets the collection of orchestration ids which should be skipped for cleanup
+    /// </summary>
+    /// <value>A set of one or more <see cref="string"/></value>
+    public IReadOnlyCollection<string>? InstancesToSkipPurging { get; set; } = Array.Empty<string>();
```
wsugarman, line 52: I am unsure if IDs is the abstraction we want here. Should this be names instead? Or perhaps even globs?

wsugarman, line 52: nit: if this is nullable, then we probably don't need the default value to be empty. I think you can leave it as `null`

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -42,6 +45,7 @@ public PurgeOrchestrationInstanceHistory(IOptions<PurgeHistoryOptions> options)
         PurgeHistoryOptions value = EnsureArg.IsNotNull(options?.Value, nameof(options));
```
wsugarman, line 45: Should we just save the `PurgeHistoryOptions` as the member here?

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -42,6 +45,7 @@ public PurgeOrchestrationInstanceHistory(IOptions<PurgeHistoryOptions> options)
...
         _statuses = EnsureArg.HasItems(value.Statuses, nameof(options));
         _minimumAge = TimeSpan.FromDays(EnsureArg.IsGt(value.MinimumAgeDays, 0, nameof(options)));
+        _instancesToSkipPurging = value.InstancesToSkipPurging == null ? new HashSet<string>() : value.InstancesToSkipPurging.ToHashSet<string>();
```
wsugarman, line 48: While often a good idea to use a `HashSet<string>` I can't imagine these collections are going to be large enough to merit it

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -65,18 +69,30 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
+        var instances = await client.ListInstancesAsync(condition, CancellationToken.None);
+
+        var instancesToPurge = instances.DurableOrchestrationState.Where(x => !_instancesToSkipPurging.Contains(x.InstanceId, StringComparer.OrdinalIgnoreCase));
 
-        if (result.InstancesDeleted > 0)
+        foreach(var instance in instancesToPurge)
         {
-            log.LogInformation("Deleted {Count} orchestration instances from storage.", result.InstancesDeleted);
+            var result = await client.PurgeInstanceHistoryAsync(instance.InstanceId);
+            log.LogInformation("Instance with {InstanceId} deleted from storage", result.InstancesDeleted);
```
wsugarman, line 91: From a log perspective, it'd be helpful to include what kind of job (i.e. name) in the log too

wsugarman, line 91: nit: `InstancesDeleted` is not the ID, is it?

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -65,18 +69,30 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
         }
-        else
-        {
+
+        if (!instancesToPurge.Any())
```
wsugarman, line 94: Should we instead perhaps count the number of instances deleted and print that to the log?

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -43,14 +46,57 @@ public PurgeOrchestrationInstanceHistoryTests()
...
+                DurableOrchestrationState = durableOrchestrationState
+            });
+
         _durableClient
-            .PurgeInstanceHistoryAsync(
-                DateTime.MinValue,
-                _utcNow.AddDays(-_purgeConfig.MinimumAgeDays),
-                _purgeConfig.Statuses)
-            .Returns(new PurgeHistoryResult(deleted));
+            .PurgeInstanceHistoryAsync(Arg.Any<string>())
```
wsugarman, line 98: nit: we should probably assert the ID passed to the method

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -43,14 +46,57 @@ public PurgeOrchestrationInstanceHistoryTests()
...
+        _purgeConfig.InstancesToSkipPurging = new string[] { instanceId1 };
+        _purgeTask = new PurgeOrchestrationInstanceHistory(Options.Create(_purgeConfig));
+
+        var durableOrchestrationState = new List<DurableOrchestrationStatus> {
+            new DurableOrchestrationStatus { InstanceId = instanceId1 },
+            new DurableOrchestrationStatus { InstanceId = instanceId2 }
+        };
+
+        _durableClient
+            .ListInstancesAsync(Arg.Any<OrchestrationStatusQueryCondition>(), Arg.Any<CancellationToken>())
```
wsugarman, line 91: nit: let's assert the query

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: We could, but then we also do some validations in the below method to check for nulls

wsugarman: We can check for nulls while just keeping the options, right?

src/Microsoft.Health.Operations.Functions/Management/PurgeHistoryOptions.cs
bcarthic: Changed to name

bcarthic: Done

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Changed to array

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: that was a mistake, updated

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Not sure If I explained what I meant. I updated the code to have options as  member variable but need to move null checks to Run method.

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -57,27 +55,39 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
         EnsureArg.IsNotNull(client, nameof(client));
         EnsureArg.IsNotNull(myTimer, nameof(myTimer));
         EnsureArg.IsNotNull(log, nameof(log));
+        IReadOnlyCollection<OrchestrationRuntimeStatus>? statuses = EnsureArg.HasItems(_options.Statuses, nameof(_options.Statuses));
```
wsugarman, line 58: Can't we assert these in the ctor? I don't think we need to assert these everytime, although if someone keeps the reference I suppose it could wreck havoc.

As an aside, we're mostly making these checks for atypical cases, as the data annotations should cover validation when using `IOptions<T>`

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -57,27 +55,39 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
+            CreatedTimeFrom = DateTime.MinValue,
+            CreatedTimeTo = end.UtcDateTime,
+            RuntimeStatus = statuses
+        };
+
+        var instances = await client.ListInstancesAsync(condition, CancellationToken.None);
+
+        var instancesToPurge = instances.DurableOrchestrationState.Where(x => !instancesToSkipPurging.Contains(x.Name, StringComparer.OrdinalIgnoreCase));
+
+        foreach(var instance in instancesToPurge)
```
wsugarman, line 85: nit: space missing after `foreach`

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -57,27 +55,39 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
         {
-            log.LogInformation("No Orchestration instances found within given parameters.");
+            var result = await client.PurgeInstanceHistoryAsync(instance.InstanceId);
+            log.LogInformation("Instance '{InstanceName}' deleted from storage", instance.Name);
```
wsugarman, line 88: Oh, I think we want the ID too haha

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -57,27 +55,39 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
         }
+
+        log.LogInformation("Deleted {Count} orchestration instances from storage.", instancesToPurge.Count());
```
wsugarman, line 91: Perhaps we can just have a local variable we increment in the loop, as opposed to enumerating through the instances again for `Count()`

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -57,27 +55,39 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
         {
-            log.LogInformation("Deleted {Count} orchestration instances from storage.", result.InstancesDeleted);
-        }
-        else
+            CreatedTimeFrom = DateTime.MinValue,
+            CreatedTimeTo = end.UtcDateTime,
+            RuntimeStatus = statuses
+        };
+
+        var instances = await client.ListInstancesAsync(condition, CancellationToken.None);
```
wsugarman, line 81: nit: this `var` type isn't obvious without intellisense

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
```diff
@@ -57,27 +55,39 @@ public async Task Run([TimerTrigger(PurgeFrequencyVariable)] TimerInfo myTimer,
...
+
+        var instancesToPurge = instances.DurableOrchestrationState.Where(x => !instancesToSkipPurging.Contains(x.Name, StringComparer.OrdinalIgnoreCase));
```
wsugarman, line 83: nit: this `var` type isn't obvious without intellisense

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
wsugarman, line 91: nit: technically this is agnostic of the back-end, so perhaps:
```suggestion
        log.LogInformation("Deleted {Count} orchestration instances from the task hub.", instancesToPurge.Count());
```

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -43,14 +46,57 @@ public PurgeOrchestrationInstanceHistoryTests()
...
+            new DurableOrchestrationStatus { InstanceId = instanceId2, Name = instanceId2 }
+        };
+
         _durableClient
-            .PurgeInstanceHistoryAsync(
-                DateTime.MinValue,
-                _utcNow.AddDays(-_purgeConfig.MinimumAgeDays),
-                _purgeConfig.Statuses)
-            .Returns(new PurgeHistoryResult(deleted));
+            .ListInstancesAsync(Arg.Is<OrchestrationStatusQueryCondition>(condition => AreConditionEqual(condition)), Arg.Any<CancellationToken>())
```
wsugarman, line 91: nit: 
```suggestion
            .ListInstancesAsync(Arg.Is<OrchestrationStatusQueryCondition>(AreConditionEqual), Arg.Any<CancellationToken>())
```

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -43,14 +46,57 @@ public PurgeOrchestrationInstanceHistoryTests()
...
+    [Fact]
+    public async Task GivenOrchestrationInstancesAndInstancesToSkipPurging_WhenPurgeCompletedDurableFunctionsHistory_ThenNoOrchestrationsPurgedAsync()
+    {
+        var instanceId1 = Guid.NewGuid().ToString();
+        var instanceId2 = Guid.NewGuid().ToString();
+
+        _purgeConfig.InstancesToSkipPurging = new string[] { instanceId1 };
+        _purgeTask = new PurgeOrchestrationInstanceHistory(Options.Create(_purgeConfig));
+
+        var durableOrchestrationState = new List<DurableOrchestrationStatus> {
```
wsugarman, line 85: nit: `{` on the next line

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -43,14 +46,57 @@ public PurgeOrchestrationInstanceHistoryTests()
...
     [InlineData(12)]
-    public async Task GivenNoOrchestrationInstances_WhenPurgeCompletedDurableFunctionsHistory_ThenNoOrchestrationsPurgedAsync(int deleted)
+    public async Task GivenOrchestrationInstances_WhenPurgeCompletedDurableFunctionsHistory_ThenOrchestrationsPurgedAsync(int count)
     {
+        var instanceId = Guid.NewGuid().ToString();
+
+        var durableOrchestrationState = Enumerable.Repeat(new DurableOrchestrationStatus { InstanceId = instanceId }, count);
+
+        _durableClient
+            .ListInstancesAsync(Arg.Is<OrchestrationStatusQueryCondition>(condition => AreConditionEqual(condition)), Arg.Any<CancellationToken>())
```
wsugarman, line 56: nit: 
```suggestion
            .ListInstancesAsync(Arg.Is<OrchestrationStatusQueryCondition>(AreConditionEqual), Arg.Any<CancellationToken>())
```

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -24,16 +27,16 @@ public class PurgeOrchestrationInstanceHistoryTests
     private readonly DateTime _utcNow;
     private readonly TimerInfo _timer;
     private readonly PurgeHistoryOptions _purgeConfig;
-    private readonly PurgeOrchestrationInstanceHistory _purgeTask;
     private readonly IDurableOrchestrationClient _durableClient;
+    private PurgeOrchestrationInstanceHistory _purgeTask;
```
wsugarman, line 31: This can still be `readonly`, right? You just need to modify the values of the `_purgeConfig` before testing

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
bcarthic: Not sure if that works, it cannot convert a method group to a condition

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: done

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Okay :)

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
bcarthic: this didnt work

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions/Management/PurgeOrchestrationInstanceHistory.cs
bcarthic: Done

wsugarman (approved):

src/Microsoft.Health.Operations.Functions/Management/PurgeHistoryOptions.cs
```diff
@@ -43,4 +44,10 @@ public class PurgeHistoryOptions
     /// <value>A value cron expression</value>
     [Required]
     public string? Frequency { get; set; }
+
+    /// <summary>
+    ///  Gets or sets the collection of orchestration names which should be skipped for cleanup
+    /// </summary>
+    /// <value>A set of one or more <see cref="string"/></value>
+    public IReadOnlyCollection<string>? InstancesToSkipPurging { get; set; }
```
wsugarman, line 52: Perhaps something a bit shorter, like `Exclude`? `ExcludeFunctions`?

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
```diff
@@ -43,14 +45,57 @@ public PurgeOrchestrationInstanceHistoryTests()
...
+    public async Task GivenOrchestrationInstancesAndInstancesToSkipPurging_WhenPurgeCompletedDurableFunctionsHistory_ThenNoOrchestrationsPurgedAsync()
+    {
+        var instanceId1 = Guid.NewGuid().ToString();
+        var instanceId2 = Guid.NewGuid().ToString();
+
+        _purgeConfig.InstancesToSkipPurging = new string[] { instanceId1 };
+
+        var durableOrchestrationState = new List<DurableOrchestrationStatus> 
+        {
+            new DurableOrchestrationStatus { InstanceId = instanceId1, Name = instanceId1 },
```
wsugarman, line 85: This test is a bit confusing because you're using the instance Id as the name. Can you change it to reflect an actual name? Otherwise it looks like the options are used to filter out instance IDs

wsugarman (changes requested):

src/Microsoft.Health.Operations.Functions/Management/PurgeHistoryOptions.cs
bcarthic: Done

src/Microsoft.Health.Operations.Functions.UnitTests/Management/PurgeOrchestrationInstanceHistoryTests.cs
bcarthic: done

wsugarman (approved):

PR #501, opened 2022-05-25
Adds schema upgrade notification

rbans96:
## Description
This PR adds schemaUpgradedNotification to be sent out whenever the schema is initialized/upgraded so that the fhir-service can initialize dictionaries when schema is not null.

## Related issues
Addresses[ #91880](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=91880)

## Testing
Tested locally

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
@@ -11,10 +11,12 @@ using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
+using MediatR;
 using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.SqlServer.Configs;
+using Microsoft.Health.SqlServer.Features.Schema.Extensions;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -288,6 +293,10 @@ private async Task ApplySchemaInternalAsync(int version, string script, bool app
         await _schemaManagerDataStore.ExecuteScriptAndCompleteSchemaVersionAsync(script, version, applyFullSchemaSnapshot, cancellationToken).ConfigureAwait(false);
 
         _logger.LogInformation("Schema migration completed successfully for the version : {Version}.", version);
+
+        // this notification is used by the fhir service to initialized the dictionaries
```
brendankowitz, line 297: Might be better to comment what this event is for and why its useful

brendankowitz (dismissed):

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
rbans96: Sure, updated the comment

feordin (approved):

PR #505, opened 2022-05-27
Add error reporting model and interface

wi-y:
## Description
We would like to be able to report errors to a destination. This PR creates an Error class and an interface that can be used by a service that reports the error to a destination.

Feature


src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// This Id can be generated by the caller, or left null.
+        /// If null, the IErrorMessageService will generate an Id.
+        /// </remarks>
+        /// </summary>
+        public string Id { get; set; }
+
+        /// <summary>
+        /// The type of error.
+        /// </summary>
+        public string ErrorType { get; set; }
```
kyclai, line 25: Do you foresee each service that writes to the Error Messaging Service will determine its own set of error types, or would it be possible to have a generic set of error types (in which case `ErrorType` can have an enum type with these known values instead)?

kyclai (approved):

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// A property bag for data which can provide additional context for the error.
+        /// </summary>
+        public IDictionary<string, object> Values { get; private set; }
+
+        public void SetValues(IDictionary<string, object> dictionary)
+        {
+            Values = dictionary;
+        }
```
brendankowitz, line 48: Should this be a constructor?

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// The type of error.
+        /// </summary>
+        public string ErrorType { get; set; }
+
+        /// <summary>
+        /// A string describing the error that occurred.
+        /// </summary>
+        public string ErrorString { get; set; }
```
brendankowitz, line 30: If these properties `ErrorString`, `ErrorTime` all existing within the class `ErrorMessage`, should we consider removing the prefix?

Also a lot of these properties sounds like those from https://docs.microsoft.com/en-us/dotnet/api/system.exception?view=net-6.0#properties but with different names, any consideration for aligning or just using Exception class?

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+        public DateTimeOffset ErrorTime { get; set; }
+
+        /// <summary>
+        /// A property bag for data which can provide additional context for the error.
+        /// </summary>
+        public IDictionary<string, object> Values { get; private set; }
+
+        public void SetValues(IDictionary<string, object> dictionary)
+        {
+            Values = dictionary;
```
brendankowitz, line 47:
```suggestion
            Values = EnsureArgs.IsNotNull(dictionary, nameof(dictionary));
```

src/Microsoft.Health.Abstractions/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public interface IErrorMessageService
+    {
+        /// <summary>
+        /// Writes a collection of ErrorMessages to a destination
+        /// </summary>
+        /// <param name="errorMessages">A collection of error message objects.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns><see cref="Task"/>Returns the successfully stored error messages.</returns>
+        Task<IEnumerable<ErrorMessage>> ReportError(IEnumerable<ErrorMessage> errorMessages, CancellationToken cancellationToken);
```
wsugarman, line 20: This should return `IAsyncEnumerable<ErrorMessage>`, if you plan to return values

wsugarman, line 20: nit: default should be provided for token: `CancellatioonToken cancellationToken = default`

wsugarman, line 20: It's a bit inconsistent with the .NET libraries, but this should probably be `ReportErrorAsync` if it returns `IAsyncEnumerable<ErrorMessage>`. Otherwise all methods that return `Task` should use the `async` suffix

wsugarman, line 20: Will the returned messages be a subset of the `errorMessages`?

src/Microsoft.Health.Abstractions/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Abstractions.Errors
+{
+    public interface IErrorMessageService
```
wsugarman, line 12: nit: missing `<summary>` tag for interface

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    public class ErrorMessage
+    {
+        /// <summary>
+        /// A unique identifier for the ErrorMessage
+        /// <remarks>
+        /// This Id can be generated by the caller, or left null.
+        /// If null, the IErrorMessageService will generate an Id.
+        /// </remarks>
+        /// </summary>
```
wsugarman, line 19: `<remarks>` are not nested in `<summary>`. They are two separate elements

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Abstractions.Errors
+{
+    public class ErrorMessage
+    {
+        /// <summary>
+        /// A unique identifier for the ErrorMessage
```
wsugarman, line 14: Property summaries use the comment format `Gets [or sets] ...`.

Eg.
```csharp
/// <summary>
/// Gets or sets a unique identifier for the message.
/// </summary>
```

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Abstractions.Errors
+{
+    public class ErrorMessage
```
wsugarman, line 11: nit: missing `<summary>` for type

src/Microsoft.Health.Abstractions/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Writes a collection of ErrorMessages to a destination
```
wsugarman, line 15: nit: "Asynchronously writes ..."

wsugarman, line 15: nit: if you truly wish to refer explicitly to the type (which you don't have to) then you should use the proper tags. Either:

```csharp
/// Asynchronously writes a collection of <see cref="ErrorMessage" /> objects to a destination.
```

OR

```csharp
/// Asynchronously writes a collection of errror messages to a destination.
```

src/Microsoft.Health.Abstractions/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        /// <param name="errorMessages">A collection of error message objects.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns><see cref="Task"/>Returns the successfully stored error messages.</returns>
```
wsugarman, line 19: Typically the comment for an asynchronous result looks like the following:

```csharp
/// <returns>A task that represents the asynchronous operation.</returns>
```

Furthermore, if the result is `Task<T>` it does on to describe the result, then a reference is made to the result

```csharp
/// <returns>
/// A task that represents the asynchronous operation.
/// The value of the <see cref="Task{TResult}.Result"/> property contains ...
/// </returns>
```
OR

```csharp
/// <returns>
/// A task that represents the asynchronous operation.
/// The value of the <c>TResult</c> parameter contains ...
/// </returns>
```

See something like [`Stream.ReadAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-6.0#system-io-stream-readasync(system-byte()-system-int32-system-int32)) as an example.

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// A unique identifier for the ErrorMessage
+        /// <remarks>
+        /// This Id can be generated by the caller, or left null.
+        /// If null, the IErrorMessageService will generate an Id.
+        /// </remarks>
+        /// </summary>
+        public string Id { get; set; }
```
wsugarman, line 20: Typically properties also contain a `<value>` element that describes the values that may be returned. While the summary may describe the semantics of the ID, the `<value>` would be like `<value>An optional identifer</value>`

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.Health.Abstractions.Errors
+{
+    public class ErrorMessage
+    {
+        /// <summary>
+        /// A unique identifier for the ErrorMessage
+        /// <remarks>
+        /// This Id can be generated by the caller, or left null.
+        /// If null, the IErrorMessageService will generate an Id.
```
wsugarman, line 17: nit:
```suggestion
        /// If null, the <see cref="IErrorMessageService" /> will generate an Id.
```

wsugarman: Although, I am unsure you want to so tightly bound this class with the service my mentioning it explicitly here

wsugarman, line 17: nit: `null` is a `langword`:
```suggestion
        /// If <see langword="null" />, the IErrorMessageService will generate an Id.
```

wsugarman (changes requested):

src/Microsoft.Health.Abstractions/Errors/ErrorMessage.cs
wi-y: Yeah it could be an Exception since there is a quite a bit of overlap. Inheriting from Exception or just adding some extension methods for Exception might achieve the goals but it is difficult to tell at this point. I would like to do some proof of concept work some sense of how Exception/extension methods will work in practice with our other repos before committing any code to this repo. Will be putting this PR on hold/abandoning it since it is not clear that these classes will be useful.

PR #507, opened 2022-06-01
Add Client authentication helpers

brandonpollett:
## Description
This PR adds helpers to setup authentication for the `Microsoft.Health.Client` child classes. 

## Related issues
Addresses [AB#92238](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/92238).

## Testing
Manually tested downstream in DicomCast

Patch


src/Microsoft.Health.Client/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+        EnsureArg.IsNotNull(serviceCollection, nameof(serviceCollection));
+        EnsureArg.IsNotNull(managedIdentityCredentialConfiguration, nameof(managedIdentityCredentialConfiguration));
+        EnsureArg.IsNotNullOrWhiteSpace(name, nameof(name));
+
+        serviceCollection.Configure<ManagedIdentityCredentialConfiguration>(name, managedIdentityCredentialConfiguration);
+
+        serviceCollection.Add(provider =>
+            {
+                IOptionsMonitor<ManagedIdentityCredentialConfiguration> options = provider.GetService<IOptionsMonitor<ManagedIdentityCredentialConfiguration>>();
+                var httpClientFactory = provider.GetService<IHttpClientFactory>();
```
jovinson-ms, line 30: Why is this credential provider the only one that uses HttpClientFactory?

brandonpollett: The `ManagedIdentityCredentialProvider` uses a `AzureServiceTokenProvider` under the hood and it's constructor requires an `IHttpClientFactory`. It looks like we can actually switch this to a `DefaultAzureCredential` and remove this need.

src/Microsoft.Health.Client/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+        EnsureArg.IsNotNull(serviceCollection, nameof(serviceCollection));
+        EnsureArg.IsNotNull(oAuth2ClientCertificateCredentialConfiguration, nameof(oAuth2ClientCertificateCredentialConfiguration));
+        EnsureArg.IsNotNullOrWhiteSpace(name, nameof(name));
+
+        serviceCollection.Configure<OAuth2ClientCertificateCredentialConfiguration>(name, oAuth2ClientCertificateCredentialConfiguration);
+
+        serviceCollection.Add(provider =>
+            {
+                IOptionsMonitor<OAuth2ClientCertificateCredentialConfiguration> options = provider.GetService<IOptionsMonitor<OAuth2ClientCertificateCredentialConfiguration>>();
+                var httpClient = new HttpClient();
```
wsugarman, line 47: There seems something weird about creating an `HttpClient` to fetch auth for another `HttpClient`. I am wondering if we should be using something like the `AddHttpMessageHandler` extension method to create our own [`DelegatingHandler`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.delegatinghandler?view=net-6.0) that wraps these providers to optionally retrieve the token if necessary.

wsugarman (changes requested):

src/Microsoft.Health.Client/Extensions/IHttpClientBuilderRegistrationExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+using EnsureThat;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.Client.Configuration;
+
+namespace Microsoft.Health.Client.Extensions;
+
+public static class IHttpClientBuilderRegistrationExtensions
+{
+    public static void AddAuthenticationHandler(this IHttpClientBuilder httpClientBuilder, IServiceCollection services, IConfigurationSection authenticationConfigurationSection, string credentialProviderName)
```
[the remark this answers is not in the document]
brandonpollett: We can make name optional, but I think I would want to make the default value a random value (probably a GUID). 

The reason being is that we're using that to resolve the credential provider and if we don't, we won't otherwise know which to select. The issue with the `httpClientBuilder.Name` is that it appears that if a name isn't specified, it's either going to be `Options.DefaultName` or the `TypeNameHelper.GetTypeDisplayName(typeof(TClient), fullName: false);` which can both have collisions.

src/Microsoft.Health.Client/Configuration/AuthenticationConfiguration.cs
```diff
@@ -0,0 +1,23 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Client.Configuration;
```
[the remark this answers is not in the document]
wsugarman: Yeah, I've been slowly trying to make new settings more "agnostic" of being configurations. I see this usage in the Microsoft extensions where settings are called "Options" and they are present in their assembly's root namespace (or at least one that isn't specific to being options/configs), like:
- [`MemoryCacheOptions`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheoptions?view=dotnet-plat-ext-6.0)
- [`HttpClientFactoryOptions`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.http.httpclientfactoryoptions?view=dotnet-plat-ext-6.0)
- [`ConsoleLoggerOptions`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.console.consoleloggeroptions?view=dotnet-plat-ext-6.0)

That being said, we have used this pattern of "Configuration" classes in the "Configuration" namespace in so many places that I am not going to stop one more example of it lol.

src/Microsoft.Health.Client/Extensions/ServiceCollectionExtensions.cs
wsugarman: Oh! Yes -- of course that's at the end of the registration. And that makes sense about needing to fetch the token.

wsugarman: Looking at this service registration, it makes sense that it is singleton, as the token can be used across multiple users' requests. However, I think it'd be ideal for the credential to be singleton (as well as thread-safe!) and long-lived while the actual client is short-lived, as there can be [DNS problems with singleton clients](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net). Perhaps we use a separate `IHttpClientFactory` for these auth clients? I am unsure if this needs to be addressed immediately, but it's food for thought.

src/Microsoft.Health.Client/ManagedIdentityCredentialProvider.cs
```diff
@@ -5,40 +5,43 @@
...
         _optionsName = optionsName;
-        _httpClientFactory = httpClientFactory;
     }
 
     protected override async Task<string> BearerTokenFunction(CancellationToken cancellationToken)
     {
         ManagedIdentityCredentialConfiguration managedIdentityCredentialConfiguration = _managedIdentityCredentialConfigurationMonitor.Get(_optionsName);
 
-        var azureServiceTokenProvider = new AzureServiceTokenProvider(httpClientFactory: _httpClientFactory);
-        return await azureServiceTokenProvider.GetAccessTokenAsync(managedIdentityCredentialConfiguration.Resource, managedIdentityCredentialConfiguration.TenantId, cancellationToken);
```
wsugarman, line 42: Coincidentally, I was just talking to Jared about this class yesterday. Should we be using `Azure.Identity` elsewhere? Like using `ClientCertificateCredential` for `OAuth2ClientCertificateCredential`? Or are we ensuring this is agnostic of Azure?

src/Microsoft.Health.Client/Extensions/IHttpClientBuilderRegistrationExtensions.cs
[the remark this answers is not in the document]
wsugarman: I am thinking for scenarios where there is only 1 set of credentials configured in this manner, it may make sense to leave it optional. And in those cases, I think the default name like `""` should be sufficient if there is only 1 configured credential provider.

src/Microsoft.Health.Client/ManagedIdentityCredentialProvider.cs
brandonpollett: We're ensuring that this is agnostic of azure to support our OSS use cases.

src/Microsoft.Health.Client/Authentication/AuthenticationOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.Health.Client.Authentication;
+
+public class AuthenticationOptions
+{
+    public const string SectionName = "Authentication";
+
+    public bool Enabled { get; set; }
+
+    public AuthenticationType? AuthenticationType { get; set; }
```
wsugarman, line 14: Do we need `AuthenticationType`  if we have options for each type below?
Eg. `if (OAuth2ClientCertificateCredential != null) { ... }`

src/Microsoft.Health.Client/Authentication/ICredentialProvider.cs
```diff
@@ -6,7 +6,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.Health.Client;
+namespace Microsoft.Health.Client.Authentication;
 
 public interface ICredentialProvider
```
wsugarman, line 11: Perhaps outside the scope of these changes, but should `GetBearerToken` be `GetBearerTokenAsync`?

src/Microsoft.Health.Client/Authentication/ManagedIdentityCredentialProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <param name="optionsName">Optional name to use when retrieving options from the IOptionsMonitor</param>
+    public ManagedIdentityCredentialProvider(IOptionsMonitor<ManagedIdentityCredentialOptions> managedIdentityCredentialOptionsMonitor, string optionsName = null)
+    {
+        EnsureArg.IsNotNull(managedIdentityCredentialOptionsMonitor, nameof(managedIdentityCredentialOptionsMonitor));
+
+        _managedIdentityCredentialOptionsMonitor = managedIdentityCredentialOptionsMonitor;
+        _optionsName = optionsName;
+    }
+
+    protected override async Task<string> BearerTokenFunction(CancellationToken cancellationToken)
```
wsugarman, line 34: I know this is from the abstract class, but I'm not wild about this function name lol

src/Microsoft.Health.Client/Authentication/OAuth2ClientCertificateCredentialProvider.cs
```diff
@@ -12,52 +12,52 @@
...
-namespace Microsoft.Health.Client;
+namespace Microsoft.Health.Client.Authentication;
 
 public class OAuth2ClientCertificateCredentialProvider : CredentialProvider
 {
     private const string JwtAssertionType = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
     private const string Rs256Algorithm = "RS256";
 
-    private readonly IOptionsMonitor<OAuth2ClientCertificateCredentialConfiguration> _oAuth2ClientCertificateCredentialConfigurationMonitor;
+    private readonly IOptionsMonitor<OAuth2ClientCertificateCredentialOptions> _oAuth2ClientCertificateCredentialOptionsMonitor;
```
wsugarman, line 26: nit: you could probably just call this `_options` lol

src/Microsoft.Health.Client/Extensions/IHttpClientBuilderRegistrationExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+            case AuthenticationType.OAuth2ClientCredential:
+                httpClientBuilder.Services.AddNamedOAuth2ClientCredentialProvider(authenticationConfigurationSection.GetSection(nameof(auth.OAuth2ClientCredential)), httpClientBuilder.Name);
+                break;
+            case AuthenticationType.OAuth2UserPasswordCredential:
+                httpClientBuilder.Services.AddNamedOAuth2UserPasswordCredentialProvider(authenticationConfigurationSection.GetSection(nameof(auth.OAuth2UserPasswordCredential)), httpClientBuilder.Name);
+                break;
+        }
+
+        httpClientBuilder
+            .AddHttpMessageHandler(x => new AuthenticationHttpMessageHandler(x.ResolveNamedCredentialProvider(httpClientBuilder.Name)));
```
wsugarman, line 44: I think you can avoid the closure here by retrieving the name from the `x` parameter:
```suggestion
            .AddHttpMessageHandler(x => new AuthenticationHttpMessageHandler(x.ResolveNamedCredentialProvider(x.Name)));
```

wsugarman (approved):

src/Microsoft.Health.Client/Authentication/AuthenticationOptions.cs
brandonpollett: I think I prefer being explicit here. My reasoning is that without being explicit we're implying some sort of order of precedence.

src/Microsoft.Health.Client/Extensions/IHttpClientBuilderRegistrationExtensions.cs
wsugarman: Oh you're right -- totally misread the doc

PR #520, opened 2022-06-28
ServiceCollection Registration updates

v-mserdar:
## Description
Added a new collection that has the same messages that the Debug.Assert functions were raising when services were not correctly registered. The point of this is to merely expose these messages that already provided guidance on how they should be used.
Revised the services.Add() to be services.TryAdd() to prevent duplicates getting added.
Added a unit test.

## Related issues
Addresses [92947](https://microsofthealth.visualstudio.com/Health/_workitems/edit/92947)

## Testing
Added a new unit test and locally successfully ran all tests in Microsoft.Health.Extensions.DependencyInjection.UnitTests (net6.0)

Patch|Skip|Feature

### DIFF
--- a/src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -4,7 +4,6 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
@@ -30,10 +29,11 @@ public static class ServiceCollectionExtensions
         EnsureArg.IsNotNull(collection, nameof(collection));
 
         var moduleType = typeof(T);
-        Debug.Assert(moduleType.IsClass && !moduleType.IsAbstract, "Module should be non-abstract class");
+        bool assertion = moduleType.IsClass && !moduleType.IsAbstract;
+        Debug.Assert(assertion, "Module should be non-abstract class");
 
         // have the registration follow same rules as fn RegisterAssemblyModules
-        if (moduleType.IsClass && !moduleType.IsAbstract)
+        if (assertion)
         {
             collection.RegisterModule(moduleType, constructorParams);
         }
@@ -55,25 +55,6 @@ public static class ServiceCollectionExtensions
         {
             collection.RegisterModule(moduleType, constructorParams);
         }
-    }
-
-    public static ICollection<string> CheckIssues(this IServiceCollection collection)
-    {
-        var issues = new List<string>();
-        var multipleRegisteredServices = collection
-            .GroupBy(x => (x.ServiceType, x.ImplementationType, x.ImplementationInstance, x.ImplementationFactory))
-            .Where(x => x.Count() > 1)
-            .ToArray();
-
-        if (multipleRegisteredServices.Any())
-        {
-            foreach (var service in multipleRegisteredServices)
-            {
-                issues.Add($"** IoC Config Warning: Service implementation '{service.Key.ImplementationType ?? service.Key.ImplementationInstance ?? service.Key.ImplementationFactory}' was registered multiple times.");
-            }
-        }
-
-        return issues;
     }
 
     private static void RegisterModule(this IServiceCollection collection, Type moduleType, params object[] constructorParams)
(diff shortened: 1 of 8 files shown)


wsugarman (changes requested):

src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -51,6 +57,25 @@ public static void RegisterAssemblyModules(this IServiceCollection collection, A
         }
     }
 
+    public static ICollection<string> CheckIssues(this IServiceCollection collection)
+    {
+        var issues = new List<string>();
+        var multipleRegisteredServices = collection
+            .GroupBy(x => (x.ServiceType, x.ImplementationType, x.ImplementationInstance, x.ImplementationFactory))
```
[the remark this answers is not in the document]
v-mserdar: Thanks for the feedback here. Based on the comments, especially the 3rd bullet, I should probably remove the code altogether. What started all of this was a local debug session of fhir-server hitting Debug.Asserts in TypeRegistrationBuilder.AsImplementedInterfaces() . The debug message pointed to missing AsSelf() in a handlful of areas and that's where I noticed some other debug code trying to identify duplicate service registration (which is what I added above). 
Long story short, I should probably revert/close this PR.
If you have a bug/workitem # for the Newtonsoft.json.dll CG issue, I'll use that for a new PR otherwise I'll create an item for it.

src/Microsoft.Health.Extensions.DependencyInjection/TypeRegistrationBuilder.cs
```diff
@@ -122,9 +135,14 @@ public TypeRegistrationBuilder ReplaceService<T>()
     /// <exception cref="NotSupportedException">Throws when Type was not explicitly defined</exception>
     public TypeRegistrationBuilder AsImplementedInterfaces(Predicate<Type> interfaceFilter = null)
     {
-        Debug.Assert(
-            _firstRegisteredType != null || _registrationMode == TypeRegistration.RegistrationMode.Transient,
-            $"Using \"AsImplementedInterfaces()\" without calling \"AsSelf()\" first in registration for \"{_type.Name}\" could have inconsistent results.");
+        bool assertion = _firstRegisteredType != null || _registrationMode == TypeRegistration.RegistrationMode.Transient;
+        string msg = $"Using \"AsImplementedInterfaces()\" without calling \"AsSelf()\" first in registration for \"{_type.Name}\" could have inconsistent results.";
```
wsugarman, line 139: Instead of warnings, should there be a setting on the `TypeRegistrationBuilder` when first created that enforces these assertions, like:
```csharp
IServiceCollection services = new ServiceCollection();
services
    .Add<T>(strictRegistration: true) // Or whatever name we like lol
    .AsSelf()
    .AsInterfaces();
```

Alternatively, this sort of detection would probably be a good candidate for a [custom Roslyn Analyzer](https://devblogs.microsoft.com/dotnet/how-to-write-a-roslyn-analyzer/)

wsugarman (changes requested):

src/Microsoft.Health.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -29,8 +29,14 @@ public static void RegisterModule<T>(this IServiceCollection collection, params
         EnsureArg.IsNotNull(collection, nameof(collection));
 
         var moduleType = typeof(T);
-        Debug.Assert(moduleType.IsClass && !moduleType.IsAbstract, "Module should be non-abstract class");
-        collection.RegisterModule(moduleType, constructorParams);
+        bool assertion = moduleType.IsClass && !moduleType.IsAbstract;
+        Debug.Assert(assertion, "Module should be non-abstract class");
+
+        // have the registration follow same rules as fn RegisterAssemblyModules
```
LTA-Thinking, line 35: Nit: Typo, fn

src/Microsoft.Health.Extensions.DependencyInjection/TypeRegistrationBuilder.cs
LTA-Thinking: Instead of a Debug.Assert could we have it throw an exception? That way it would be still be caught by components downstream. I agree with will a strictMode or similar setting would be useful, but by default set it to be enabled. If there is a case where we know we need the "wrong" behavior we can override the exception check.

jovinson-ms:
Closed after confirming with @v-mserdar.

PR #523, opened 2022-07-01
Throw more specific schema exceptions

brandonpollett:
## Description
Throws more specific exceptions when instance schema information is not found in the DB.

## Related issues
Addresses [AB#93113](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/93113)

## Testing
Manual testing downstream.

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer.Api.UnitTests/Features/CompatibilityVersionHandlerTests.cs
+++ b/src/Microsoft.Health.SqlServer.Api.UnitTests/Features/CompatibilityVersionHandlerTests.cs
@@ -7,7 +7,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using MediatR;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Health.Extensions.DependencyInjection;
 using Microsoft.Health.SqlServer.Api.Features;
 using Microsoft.Health.SqlServer.Features.Schema;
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.SqlServer.Api.UnitTests/Features/CompatibilityVersionHandlerTests.cs
```diff
@@ -29,7 +28,7 @@ public CompatibilityVersionHandlerTests()
     {
         _schemaMigrationDataStore = Substitute.For<ISchemaDataStore>();
         var collection = new ServiceCollection();
-        collection.Add(sp => new CompatibilityVersionHandler(_schemaMigrationDataStore)).Singleton().AsSelf().AsImplementedInterfaces();
+        collection.Add(_ => new CompatibilityVersionHandler(_schemaMigrationDataStore)).Singleton().AsSelf().AsImplementedInterfaces();
```
wsugarman, line 31: I don't know if this does anything lol. Maybe it's changed, but this now might be a variable called `_`. Although, this totally may be a discard now in recent C# versions

wsugarman (approved):

src/Microsoft.Health.SqlServer.Api.UnitTests/Features/CompatibilityVersionHandlerTests.cs
brandonpollett: Yea... just found this documentation "For backwards compatibility, if only a single input parameter is named _, then, within a lambda expression, _ is treated as the name of that parameter.". I still think that it is slightly more readable in that we don't expect to use this passed in variable.

wsugarman (approved):

PR #527, opened 2022-07-05
Add output parameters support

tongwu-msft:
## Description
Add output parameters support

## Related issues
Addresses [issue #].

## Testing

Patch|Skip|Feature|Breaking


tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -248,6 +255,42 @@ private MemberDeclarationSyntax AddPopulateCommandMethodForTableValuedParameters
                                     .WithNameColon(NameColon(ParameterNameForParameter(p)))).ToArray())));
     }
 
+    private MethodDeclarationSyntax[] AddGetOutputMethod(CreateOrAlterProcedureStatement node)
```
wsugarman, line 258: While I'm kind of bummed to see this method pretty much duplicated, it seems that we already do this today with the other methods. So it's fine

src/Microsoft.Health.SqlServer/Features/Schema/Model/ParameterDefinition.cs
```diff
@@ -57,8 +57,12 @@ public virtual SqlParameter AddParameter(SqlParameterCollection parameters, T va
     {
         EnsureArg.IsNotNull(parameters, nameof(parameters));
 
-        return parameters.Add(
-            new SqlParameter(
+        return parameters.Add(CreateSqlParameter(value));
+    }
+
+    protected virtual SqlParameter CreateSqlParameter(T value)
```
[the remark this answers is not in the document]
wsugarman: Actually -- should `OutputParameter` even be derived from `ParameterDefinition`? It's a bit weird that `AddParameter(T)` exists for the `OutputParameter` as it accepts a value which output parameters do not use. Would it be better to re-use the existing `ParameterDefintition<T>` for both input and output parameters, like `SqlParameter` does in the SQL lib? I think there are probably 2 options:
1. Separate parameter definition classes that have separate `CreateSqlParameter` method signatures such that the output parameter doesn't accept a value
2. Re-use the same type for both input and output parameters by providing new constructor(s) that accept the parameter direction (and maybe a `CreateSqlParameter` overload that doesn't accept a value).

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -248,6 +255,42 @@ private MemberDeclarationSyntax AddPopulateCommandMethodForTableValuedParameters
...
+            if (parameter.Modifier != ParameterModifier.Output)
+            {
+                continue;
+            }
+
+            TypeSyntax typeSyntax = DataTypeReferenceToClrType(parameter.DataType, parameter.Value != null);
+            result.Add(
+                MethodDeclaration(
+                    typeSyntax,
+                    Identifier($"Get{ParameterNameForParameter(parameter)}FromOutput"))
```
wsugarman, line 273: If I am reading this correctly, a user would do something like the following, correct?
```csharp
long result = VLatest.Output1.GetResultFromOutput(command);
```
If a user has multiple outputs, I think we should probably return them all in one call, perhaps through a tuple if there is more than one:
```csharp
// One output parameter
long result = VLatest.Output1.GetOutput(command);

// n output parameters
(long result1, long result2) = VLatest.Output1.GetOutput(command);
```

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateProcedureVisitor.cs
```diff
@@ -63,9 +63,12 @@ public override void Visit(CreateProcedureStatement node)
 
                 // add fields for each parameter
                 .AddMembers(node.Parameters.Select(CreateFieldForParameter).ToArray())
-
+
```
wsugarman, line 66: nit: extra white space

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -248,6 +255,42 @@ private MemberDeclarationSyntax AddPopulateCommandMethodForTableValuedParameters
                                     .WithNameColon(NameColon(ParameterNameForParameter(p)))).ToArray())));
     }
 
+    private MethodDeclarationSyntax[] AddGetOutputMethod(CreateOrAlterProcedureStatement node)
+    {
+        List<MethodDeclarationSyntax> result = new List<MethodDeclarationSyntax>();
+
+        foreach (ProcedureParameter parameter in node.Parameters)
+        {
+            if (parameter.Modifier != ParameterModifier.Output)
```
wsugarman, line 264: nit: can include in the above `foreach` as a `Where` predicate: `node.Parameters.Where(p => p.Modifier == ParameterModifier.Output)`

wsugarman (changes requested):

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
tongwu-msft: Good idea, updated. Thanks

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateProcedureVisitor.cs
tongwu-msft: thanks, updated.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
tongwu-msft: Thanks, updated.

src/Microsoft.Health.SqlServer/Features/Schema/Model/ParameterDefinition.cs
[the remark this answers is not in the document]
tongwu-msft: For OutputParameterDefintition it has special method GetOutput, looks like it should not in the ParameterDefintition. I also think that is possible some changes in ParameterDefintition, we want OutputParameterDefintition do not need to update.

Agree that we can use Direction as a virtual property.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -248,6 +255,50 @@ private MemberDeclarationSyntax AddPopulateCommandMethodForTableValuedParameters
                                     .WithNameColon(NameColon(ParameterNameForParameter(p)))).ToArray())));
     }
 
+    private MethodDeclarationSyntax[] AddGetOutputMethod(CreateOrAlterProcedureStatement node)
+    {
+        List<TypeSyntax> outputTypes = new List<TypeSyntax>();
+        List<ExpressionSyntax> outputExpressions = new List<ExpressionSyntax>();
+        IEnumerable<ProcedureParameter> procedureParameters = node.Parameters.Where(p => p.Modifier == ParameterModifier.Output);
```
wsugarman, line 262: nit: Let's call `ToList()` here so you can easily call `.Count` below

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -248,6 +255,50 @@ private MemberDeclarationSyntax AddPopulateCommandMethodForTableValuedParameters
...
+        if (procedureParameters.Count() == 0)
+        {
+            return new MethodDeclarationSyntax[0];
+        }
+
+        foreach (ProcedureParameter parameter in procedureParameters)
+        {
+            outputTypes.Add(DataTypeReferenceToClrType(parameter.DataType, parameter.Value != null));
+            outputExpressions.Add(InvocationExpression(
+                                    MemberAccessExpression(
```
wsugarman, line 273: nit: no need to align these arbitrarily based on the length of identifiers. Simply indent it once per level:
```csharp
outputExpressions.Add(
    InvocationExpression(
        MemberAccessExpression(
            SyntaxKind.SimpleMemberAccessExpression,
            IdentifierName(FieldNameForParameter(parameter)),
            IdentifierName("GetOutputValue")))
    .AddArgumentListArguments(
        Argument(IdentifierName(CommandParameterName))));
```

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -248,6 +255,50 @@ private MemberDeclarationSyntax AddPopulateCommandMethodForTableValuedParameters
                                     .WithNameColon(NameColon(ParameterNameForParameter(p)))).ToArray())));
     }
 
+    private MethodDeclarationSyntax[] AddGetOutputMethod(CreateOrAlterProcedureStatement node)
+    {
+        List<TypeSyntax> outputTypes = new List<TypeSyntax>();
+        List<ExpressionSyntax> outputExpressions = new List<ExpressionSyntax>();
+        IEnumerable<ProcedureParameter> procedureParameters = node.Parameters.Where(p => p.Modifier == ParameterModifier.Output);
+
+        if (procedureParameters.Count() == 0)
```
wsugarman, line 264: Let's special-case `Count == 1` to simply return the type `T` instead of `ValueTuple<T>`

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
```diff
@@ -94,7 +97,11 @@ private MemberDeclarationSyntax CreateFieldForParameter(ProcedureParameter param
         if (TryGetSqlDbTypeForParameter(parameter, out SqlDbType sqlDbType))
         {
             // new ParameterDefinition<int>("@paramName", SqlDbType.Int, nullable, maxlength,...)
-            typeName = GenericName("ParameterDefinition")
+            string parameterDefinitionType = parameter.Modifier != ParameterModifier.Output 
+                                                ? "ParameterDefinition"
```
wsugarman, line 101: nit: indent these only 1 level

tongwu-msft: thanks, updated.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
tongwu-msft: Thanks, updated.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateOrAlterProcedureVisitor.cs
tongwu-msft: Thanks, updated.

wsugarman (approved):

PR #543, opened 2022-08-07
Operation API Refactors

wsugarman:
## Description
- Rename `Completed` to `Succeeded` to align with names in [Microsoft's REST API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#13-long-running-operations)
- Add new methods for retrieving the generic `OperationStatus` from `OrchestrationRuntimeStatus`
- Create a better API for exposing operation-specific metadata in the API response with the new `IOrchestrationCheckpoint` interface and its `GetResults` method

## Related issues
N/A

## Testing
New and old tests

Feature


src/Microsoft.Health.Operations/OperationState.cs
```diff
@@ -58,8 +58,14 @@ public class OperationState<T>
...
-    /// <value>Zero or more key-value pairs based on the <see cref="Type"/>.</value>
-    public IReadOnlyDictionary<string, string>? AdditionalProperties { get; init; }
+    /// <remarks>
+    /// The results may change over time as the operation continues execution.
+    /// </remarks>
+    /// <value>
+    /// An object whose type depends on the <see cref="OperationState{T}.Type"/> if specified;
+    /// otherwise <see langword="null"/>.
+    /// </value>
+    public object? Results { get; init; }
```
wsugarman, line 70: How do people like `Results`? Other options could be `Result`, `Output`, `Metadata`, or something else?

E.g.
```json
{
    "CreatedTime": "8/7/2022 01:23:45Z",
    "Status": "Running",
    "PercentComplete": 42,
    "Results": {
        "Exported": 1000,
        "Failed": 2,
        "ErrorHref": "foo"
    }
}
```

src/Microsoft.Health.Operations/OperationStatus.cs
```diff
@@ -28,7 +32,13 @@ public enum OperationStatus
     /// <summary>
     /// Specifies a status where the operation has finished successfully.
     /// </summary>
-    Completed,
+    Succeeded,
+
+    /// <summary>
+    /// Specifies a status where the operation has finished successfully.
+    /// </summary>
+    [Obsolete("Please use Succeeded instead to align with Microsoft's REST API guidelines.")]
```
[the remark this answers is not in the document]
wsugarman: I need to keep it for v1 of our API to avoid making a breaking change

src/Microsoft.Health.Operations/OperationState.cs
wsugarman: For export, I was thinking of renaming `"Failed"` to `"Skipped"`, but always keeping it present in the response even if it's `0`.

When you say `OperationStatus`, are you referring to the enumeration? Could you give an example of what you mean? For clarity's sake, the above example would be a sample response to a use that queried the state of their operation via `GET /operations/{operation-id}`

src/Microsoft.Health.Operations.Functions.UnitTests/DurableTask/OrchestrationRuntimeStatusExtensionsTests.cs
```diff
@@ -39,12 +40,40 @@ public void GivenStatus_WhenCheckingIfStopped_ThenReturnProperValue(Orchestratio
...
+    [InlineData(OperationStatus.Unknown, OrchestrationRuntimeStatus.Unknown)]
+    [InlineData(OperationStatus.Running, OrchestrationRuntimeStatus.Running)]
+    [InlineData(OperationStatus.Succeeded, OrchestrationRuntimeStatus.Completed)]
+#pragma warning disable CS0618, xUnit1025
+    [InlineData(OperationStatus.Completed, OrchestrationRuntimeStatus.Completed)]
+#pragma warning restore CS0618, xUnit1025
+    [InlineData(OperationStatus.Failed, OrchestrationRuntimeStatus.Failed)]
+    [InlineData(OperationStatus.Canceled, OrchestrationRuntimeStatus.Canceled)]
+    [InlineData(OperationStatus.NotStarted, OrchestrationRuntimeStatus.Pending)]
+    public void GivenStatus_WhenConvertingToOrchestrationRuntimeStatus_ThenReturnCorrespondingValue(OperationStatus status, OrchestrationRuntimeStatus expected)
```
jovinson-ms, line 63: `Status` in the Given clauses should probably be disambiguated in these tests.

jovinson-ms (dismissed):

src/Microsoft.Health.Operations.UnitTests/OperationStatusExtensionsTests.cs
```diff
@@ -23,7 +23,7 @@ public void GivenStatus_WhenCheckingIfInProgress_ThenReturnProperValue(Operation
     [InlineData(OperationStatus.Unknown, false)]
     [InlineData(OperationStatus.NotStarted, false)]
     [InlineData(OperationStatus.Running, false)]
-    [InlineData(OperationStatus.Completed, true)]
```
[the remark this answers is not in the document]
wsugarman: Kept in the test

src/Microsoft.Health.Operations/OperationStatus.cs
```diff
@@ -39,4 +42,9 @@ public enum OperationStatus
     /// Specifies a status where the operation has stopped prematurely after being told to do so by the user.
     /// </summary>
     Canceled,
+
+    /// <summary>
+    /// Specifies a status where the operation has successfully completed execution.
+    /// </summary>
+    Succeeded,
```
wsugarman, line 49: I have actually created a new value as it was impossible to serialize the value `"Completed"` for backwards compatibility.

src/Microsoft.Health.Operations/OperationState.cs
wsugarman: I think our hands are a little tied on some of the structure according to the [REST API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#percent-complete). Their structure appears to be flat for the generic operation properties
```json
{
  "createdDateTime": "2015-06-19T12-01-03.45Z",
  "percentComplete": "50",
  "status": "running"
}
```

However, I think our options become whether to include operation-specific metadata flat at the top-level or nest it under some top-level property, as the specification does not explicitly talk about it. I am leaning towards encapsulating the operation-specific metadata in a single property to neatly organize it.

wsugarman: Although maybe in the most RESTful world we would have no operation-specific metadata and instead that metadata would be made manifest through another resource. Like you could `GET` an export from like `/export/<id>` to retrieve metadata

wsugarman: Based on our discussions, an additional resource seems unnecessary at this time

jovinson-ms (approved):

PR #560, opened 2022-08-26
Implement the three common constructors for Exception classes

jgador:
## Description
Implement the three common constructors for user-defined Exception classes.

## Related issues
NA

## Testing
NA

None


wsugarman:
Thank you for the contribution @jgador. However, is there a use-case you are addressing with this pull request? Otherwise, I am reluctant to accept the changes. An analyzer rule may flag these exceptions today as not having implemented the proper exception ctors, but I think that would also include the one related to serialization.

jgador:
I don't have particular use-case aside from trying to follow the convention of user-defined exceptions that can be found here: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions.

You may close this if there is no added value as of the moment for implementing the minimum constructors for these exceptions.

PR #582, opened 2022-09-27
Update Sql Parser and Column Name

ccfromms:
## Description

- We find that TSql140Parser does not support function like string_agg within group(order by), so we try to update DacFx and change TSql140Parser to TSql150Parser. 
- We notice that if we define sql table column name like TableName, it will conflict with TableName in Table, so we try to convert such column name to another one.

## Testing
We add a stored procedure which use the function string_agg within group(order by), and test whether this function can be identified.


tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
```diff
@@ -1,8 +1,12 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
+    <TargetFrameworks>net462</TargetFrameworks>
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
+  <ItemGroup>
+    <None Remove="Sql\sprc_string_agg.sql" />
```
wsugarman, line 8: Why is this removed?

tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
```diff
@@ -1,8 +1,12 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
+    <TargetFrameworks>net462</TargetFrameworks>
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
```
wsugarman, line 5: This should be inherited via `Directory.Build.props`

tools/Microsoft.Health.Tools.Sql.Tasks/Helpers/SqlScriptWriter.cs
```diff
@@ -22,7 +22,7 @@ public SqlScriptWriter(string path)
         Directory.CreateDirectory(Path.GetDirectoryName(path));
 
         _writer = File.CreateText(path);
-        _sqlScriptGenerator = new Sql140ScriptGenerator();
+        _sqlScriptGenerator = new Sql160ScriptGenerator();
```
wsugarman, line 25: I take it this refers to SQL Server 16 which is 2022 and I think is in preview. Could we use the current production version 15 instead? Is there a `Sql150ScriptGenerator`?

tools/Microsoft.Health.Tools.Sql.Tasks/Microsoft.Health.Tools.Sql.Tasks.csproj
```diff
@@ -1,14 +1,14 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net462</TargetFramework>
```
wsugarman, line 4: Again, we don't want to use .NET Framework, right? Alternatively we could multi-target if we needed to, but I see this package has three targets: https://www.nuget.org/packages/Microsoft.SqlServer.DacFx/160.6296.0

wsugarman (changes requested):

tools/Microsoft.Health.Tools.Sql.Tasks/Microsoft.Health.Tools.Sql.Tasks.csproj
wsugarman: Also, what's the difference between this package and the old one?

Directory.Packages.props
```diff
@@ -1,5 +1,4 @@
 <Project>
-
   <ItemGroup>
```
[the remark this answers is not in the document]
ccfromms: Sorry, I will modify this file and correspond with main branch.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
```diff
@@ -22,6 +22,8 @@ namespace Microsoft.Health.Extensions.BuildTimeCodeGenerator.Sql;
 /// </summary>
 public abstract class SqlVisitor : TSqlFragmentVisitor
 {
+    private static readonly ISet<string> Keywords = new HashSet<string>() { "TableName" };
```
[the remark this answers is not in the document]
ccfromms: Ok, I will replace ISet with HashSet.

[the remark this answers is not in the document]
ccfromms: TableName is not reserved, but in [Table](https://github.com/ccfromms/healthcare-shared-components/blob/main/src/Microsoft.Health.SqlServer/Features/Schema/Model/Table.cs), we have already defined a property named TableName. If I want to set a column named TableName in a certain sql table, it will throw an exception since TableName has been used in class Table.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
```diff
@@ -266,7 +273,7 @@ protected static MemberDeclarationSyntax CreatePropertyForTableColumn(ColumnDefi
 
         return FieldDeclaration(
                 VariableDeclaration(typeName)
-                    .AddVariables(VariableDeclarator($"{column.ColumnIdentifier.Value}")
+                    .AddVariables(VariableDeclarator(GetColunmName(column.ColumnIdentifier.Value))
```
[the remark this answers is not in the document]
ccfromms: In FHIR server, we try to define a table which has a column named TableName, and TableName is conflict with the property in class [Table](https://github.com/ccfromms/healthcare-shared-components/blob/main/src/Microsoft.Health.SqlServer/Features/Schema/Model/Table.cs). Therefor we think that it is more appropriate to transform TableName here.

tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
ccfromms: Sorry, I will remove this line.

tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
ccfromms: Sorry, I will check and fix this lines.

tools/Microsoft.Health.Tools.Sql.Tasks/Helpers/SqlScriptWriter.cs
ccfromms: Ok, I will try and verify Sql150ScriptGenerator.

tools/Microsoft.Health.Tools.Sql.Tasks/Helpers/SqlScriptParser.cs
```diff
@@ -31,7 +31,7 @@ public static TSqlFragment ParseSqlFile(string sqlFile, TaskLoggingHelper taskLo
         using (var stream = File.OpenRead(sqlFile))
         using (var reader = new StreamReader(stream))
         {
-            var parser = new TSql140Parser(true);
+            var parser = new TSql160Parser(true);
```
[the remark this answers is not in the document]
ccfromms: Sorry, I will fix it in next commit.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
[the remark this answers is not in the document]
ccfromms: Ok, I will modify the name.

tools/Microsoft.Health.Tools.Sql.Tasks/Sql.targets
```diff
@@ -28,6 +28,6 @@
             SqlScript="@(SqlScript)" />
     </Target>
 
-    <UsingTask TaskName="GenerateFullScript" AssemblyFile="netstandard2.0\Microsoft.Health.Tools.Sql.Tasks.dll" />
+    <UsingTask TaskName="GenerateFullScript" AssemblyFile="netstandard2.1\Microsoft.Health.Tools.Sql.Tasks.dll" />
```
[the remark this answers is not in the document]
ccfromms: Sorry, I tried it again and found that MSBuild cannot support .NET Standard2.1. I will modify it in next commit.

tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
```diff
@@ -27,9 +28,12 @@
     <GeneratedFullScriptPath>Schema\$(LatestSchemaVersion).sql</GeneratedFullScriptPath>
   </PropertyGroup>
 
-  <Target Name="PublishSqlTasks" BeforeTargets="CoreCompile">
-    <MSBuild Projects="..\Microsoft.Health.Tools.Sql.Tasks\Microsoft.Health.Tools.Sql.Tasks.csproj" Targets="Publish" Properties="TargetFramework=netstandard2.0" />
+  <Target Name="PublishSqlTasks" BeforeTargets="CoreCompile" Condition=" '$(MSBuildRuntimeType)' == 'Core' ">
```
wsugarman, line 31: Alternative you could use a [`<choose>`](https://learn.microsoft.com/en-us/visualstudio/msbuild/choose-element-msbuild?view=vs-2022) element.

tools/Microsoft.Health.Tools.Sql.Tasks/Sql.targets
```diff
@@ -28,6 +28,7 @@
             SqlScript="@(SqlScript)" />
     </Target>
 
-    <UsingTask TaskName="GenerateFullScript" AssemblyFile="netstandard2.0\Microsoft.Health.Tools.Sql.Tasks.dll" />
+    <UsingTask TaskName="GenerateFullScript" AssemblyFile="netstandard2.1\Microsoft.Health.Tools.Sql.Tasks.dll" Condition=" '$(MSBuildRuntimeType)' == 'Core' "/>
```
wsugarman, line 31: Probably more appropriate to check would be `'$(MSBuildRuntimeType)' == 'Full'` as I think .NET Standard 2.1 should work with .NET Core/.NET or Mono

wsugarman (approved):

tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
ccfromms: Ok, I will change this element.

tools/Microsoft.Health.Tools.Sql.Tasks/Sql.targets
ccfromms: Thanks, I will fix it now.

tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
ccfromms: The choose tag should be the same level like target as this [link ](https://stackoverflow.com/questions/58090751/the-element-when-benerath-element-choose-is-unrecognized)said, so I think it will be better to use condition here.

tools/Microsoft.Health.Tools.Sql.Tasks/Sql.targets
ccfromms: I have tried it according to [MSBuildRuntimeType ](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=vs-2022#:~:text=Full%20indicating%20that%20MSBuild%20is%20running%20on%20the%20desktop%20.NET%20Framework%2C%20Core%20indicating%20that%20MSBuild%20is%20running%20on%20.NET%20Core%20(for%20example%20in%20dotnet%20build)%2C%20or%20Mono%20indicating%20that%20MSBuild%20is%20running%20on%20Mono.)and found out that it will work if we use 'Full' to specify .NET Framework and use 'Core' and 'Mono' to specify .NET Core and Mono.

PR #592, opened 2022-10-11
Fix multi-target in test.csproj

ccfromms:
## Description
The prior PR failed in CI build. I found that the netstandard2.1 folder did not build in publish and it caused the error. We restored this error in local environment and checked the log, found that the target name should be different from another one to avoid conflict. Thus, we try to rename the target in test.csproj.

## Related issues
Addresses [issue #582].

Patch|Skip|Feature|Breaking

### DIFF
--- a/tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
+++ b/tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
@@ -28,10 +28,8 @@
     <GeneratedFullScriptPath>Schema\$(LatestSchemaVersion).sql</GeneratedFullScriptPath>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


tools/Microsoft.Health.Tools.Sql.Tasks.Tests/Microsoft.Health.Tools.Sql.Tasks.Tests.csproj
```diff
@@ -28,10 +28,10 @@
     <GeneratedFullScriptPath>Schema\$(LatestSchemaVersion).sql</GeneratedFullScriptPath>
   </PropertyGroup>
 
-  <Target Name="PublishSqlTasks" BeforeTargets="CoreCompile" Condition=" '$(MSBuildRuntimeType)' == 'Core' Or '$(MSBuildRuntimeType)' == 'Mono'">
+  <Target Name="PublishSqlTasksCoreOrMono" BeforeTargets="CoreCompile" Condition=" '$(MSBuildRuntimeType)' == 'Core' Or '$(MSBuildRuntimeType)' == 'Mono'">
```
wsugarman, line 31: Alternatively, you can name them the same thing but wrap them with a choose element

wsugarman (approved):

PR #619, opened 2022-11-09
[Work in Progress] SqlException State 35

alishahahmed:
## Description
Added retry logic on Sql connection OpenAsync method when the method throws a SqlException with State 35 to take care of the following exception:

"A connection was successfully established with the server, but then an error occurred during the login process. (provider: TCP Provider, error: 35 - An internal exception was caught)"

## Related issues
[AB#96340](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/96340)

Skip

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
@@ -9,8 +9,10 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
 using Microsoft.Health.SqlServer.Configs;
 using Microsoft.Health.SqlServer.Features.Storage;
+using Polly;
 
 namespace Microsoft.Health.SqlServer.Features.Client;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -78,6 +87,34 @@ internal async Task InitializeAsync(string initialCatalog = null, CancellationTo
...
+    /// "A connection was successfully established with the server, but then an error occurred during the login process.
+    /// (provider: TCP Provider, error: 35 - An internal exception was caught)"
+    /// </remarks>
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation Token.</param>
+    private async Task OpenRetriableSqlConnectionAsync(CancellationToken cancellationToken)
+    {
+        int attempts = 1;
+
+        await Policy.Handle<SqlException>(se => se.State == 35)
```
wsugarman, line 103: nit: no need to await this. Simply return the `Task`

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -78,6 +87,34 @@ internal async Task InitializeAsync(string initialCatalog = null, CancellationTo
...
+        .WaitAndRetryAsync(
```
wsugarman, line 104: nit: indent by 1 here and below

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -78,6 +87,34 @@ internal async Task InitializeAsync(string initialCatalog = null, CancellationTo
...
+            retryCount: RetryAttempts,
+            sleepDurationProvider: (retryCount) => RetrySleepDuration,
+            onRetry: (exception, retryCount) =>
+            {
+                _logger.LogWarning(
+                    exception,
+                    "Attempt {Attempt} of {MaxAttempts} to open Sql connection.",
+                    attempts++,
```
wsugarman, line 112: Doesn't the `retryCount` parameter provide this?

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -78,6 +87,34 @@ internal async Task InitializeAsync(string initialCatalog = null, CancellationTo
         }
     }
 
+    /// <summary>
+    /// Open sql connection which is retried up to 3 times if <see cref="SqlException"/> is caught with state 35.
+    /// <remarks>
+    /// Retries have been added to cater to the following exception:
+    /// "A connection was successfully established with the server, but then an error occurred during the login process.
+    /// (provider: TCP Provider, error: 35 - An internal exception was caught)"
```
wsugarman, line 95: Are we sure this is something we want to retry? The SQL Client library already retries a number of different errors? Do we know why this isn't retried? Is this a "retryable" exception? For example, I see an old issue in the issues page for this problem when moving from .NET 5 -> 6: https://github.com/dotnet/SqlClient/issues/1402. Unsure if it's related, but I am reluctant to add new retries if we aren't quite sure why they're happening.

wsugarman (changes requested):

alishahahmed:
Closing as this will be taken care of by Daniel's work: https://microsofthealth.visualstudio.com/Health/_workitems/edit/96670.

PR #627, opened 2022-11-16
Adds more SQL retry errors

Tom-Markoch:
## Description
Adds more sql retry errors on top of the default ones.

## Testing
Does not add new functionality. Tested manually that exiting functionality still works, by injecting RAISERROR with error number being one in the list into the sql text, and observing in the log 5 retries with approx (due to randomization) exponential delay times.

### DIFF
--- a/src/Microsoft.Health.SqlServer/Configs/SqlClientRetryOptions.cs
+++ b/src/Microsoft.Health.SqlServer/Configs/SqlClientRetryOptions.cs
@@ -26,6 +26,16 @@ public class SqlClientRetryOptions
         DeltaTime = TimeSpan.FromSeconds(1),
         MaxTimeInterval = TimeSpan.FromSeconds(20),
 
+        // TODO: The following setting of the TransientErrors is temporary until the transient error retry functionality in Microsoft.Data.SqlClient is complete.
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.SqlServer/Configs/SqlClientRetryOptions.cs
```diff
@@ -24,5 +25,34 @@ public class SqlClientRetryOptions
         NumberOfTries = 5,
         DeltaTime = TimeSpan.FromSeconds(1),
         MaxTimeInterval = TimeSpan.FromSeconds(20),
+
+        // Default errors copied from src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryFactory.cs .
+        TransientErrors = new HashSet<int>
```
wsugarman, line 30: Aren't these being included by default if errors are left unspecified? So we don't need to explicit specify these values.

wsugarman (changes requested):

src/Microsoft.Health.SqlServer/Configs/SqlClientRetryOptions.cs
Tom-Markoch: Yes, but there's no way to add more, you can only replace with new set of errors. Everything is internal, sealed, protected, private.

wsugarman: Oh, I see. Sorry for the misunderstanding. I totally missed that

wsugarman (approved):

PR #634, opened 2022-12-02
Simplify blob health check

smithago:
## Description
Change blob health check from upload and download of a file to just get container properties.
This should be enough to confirm the health of blob service.
This should also reduce the memory usage. 

## Related issues
AB#98918

## Testing
Dicom blob health check continues to work


src/Microsoft.Health.Blob/Features/Storage/BlobClientContainerGetTestProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+public class BlobClientContainerGetTestProvider : IBlobClientTestProvider
+{
+    /// <inheritdoc />
+    public async Task PerformTestAsync(BlobServiceClient client, BlobContainerConfiguration blobContainerConfiguration, CancellationToken cancellationToken = default)
+    {
+        EnsureArg.IsNotNull(client, nameof(client));
+        EnsureArg.IsNotNull(blobContainerConfiguration, nameof(blobContainerConfiguration));
+
+        BlobContainerClient containerClient = client.GetBlobContainerClient(blobContainerConfiguration.ContainerName);
+        await containerClient.GetPropertiesAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
```
wsugarman, line 26: nit: you can simply return the resulting `Task` from `GetPropertiesAsync`

wsugarman (approved):

PR #637, opened 2022-12-07
[!!! Do not merge !!!] How retries work

SergeyGaluzo:
This PR shows problems with current retry logic.

### DIFF
--- a/test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/RetryTests.cs
+++ b/test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/RetryTests.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/RetryTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            // same stored procedure as above just reader added
+            await ResetError();
+            Assert.False(await ErrorIsLogged());
+            var results = await ExecuteQuery();
+            Assert.True(await ErrorIsLogged());
+            Assert.Equal(10, results.Count);
+
+            await ResetError();
+            Assert.False(await ErrorIsLogged());
+            // if error happens in select it is not handled
```
danielendeg, line 50: Is it because the transient error a different one than the default list?

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/RetryTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+            catch (SqlException ex)
+            {
+                Assert.Equal(RetryError, ex.Number);
+            }
+            Assert.True(await ErrorIsLogged());
+
+            await ResetError();
+            Assert.False(await ErrorIsLogged());
+            // if error happens after select it is not handled
```
danielendeg, line 65: Same question here, is the error code not part of the default transient errors list?

SergeyGaluzo: This exception is in the modified list. Code sees this exception. It retries it on certain cases but not all.

test/Microsoft.Health.SqlServer.Tests.Integration/Features/Schema/RetryTests.cs
SergeyGaluzo: See full answer on other comment.

PR #689, opened 2023-03-21
Types registration uses TryAdd and TryAddEnumerable

brendankowitz:
## Description
Type registration uses TryAdd and TryAddEnumerable

## Testing
Adds unit tests

Patch

### DIFF
--- a/src/Microsoft.Health.Extensions.DependencyInjection.UnitTests/TypeRegistrationTests.cs
+++ b/src/Microsoft.Health.Extensions.DependencyInjection.UnitTests/TypeRegistrationTests.cs
@@ -377,6 +377,54 @@ public class TypeRegistrationTests
         IComponent instance = componentFactory.Invoke();
 
         Assert.IsType<ComponentB>(instance);
+
+        Assert.Equal(2, provider.GetServices<IComponent>().Count());
+    }
+
+    [Fact]
+    public void GivenATypeBuilder_WhenRegisteringMultipleTimes_ThenNoDuplicatesAreAdded()
+    {
+        _collection.Add<ComponentA>()
+            .Transient()
+            .AsSelf()
+            .AsService<IComponent>();
+
+        _collection.Add<ComponentA>()
+            .Transient()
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Extensions.DependencyInjection.UnitTests/TypeRegistrationTests.cs
```diff
@@ -377,6 +377,54 @@ public void GivenADelegateFromTypeBuilder_WhenResolvingComponent_ThenResolverRet
...
+            .AsSelf()
+            .AsService<IComponent>();
+
+        var provider = _collection.BuildServiceProvider();
+        var services = provider.GetServices<IComponent>();
+        Assert.Single(services);
+    }
+
+    [Fact]
+    public void GivenATypeBuilder_WhenRegisteringMultipleTimes_ThenErrorWithDifferentScopes()
```
brendankowitz, line 403: ☝️ fix text name

src/Microsoft.Health.Extensions.DependencyInjection/TypeRegistrationBuilder.cs
```diff
@@ -255,16 +248,28 @@ private void SetupRootRegistration(Type serviceType, bool replace)
         {
             serviceDescriptor = serviceDescriptor.WithMetadata(_type);
 
-            if (replace)
-            {
-                _serviceCollection.Replace(serviceDescriptor);
```
brendankowitz, line 260: Problem with replace is that it only removes the first match it finds

src/Microsoft.Health.Extensions.DependencyInjection.UnitTests/TypeRegistrationTests.cs
brendankowitz:
```suggestion
    public void GivenATypeBuilder_WhenRegisteringMultipleTimes_ThenSecondRegistrationIsIgnored()
```

PR #696, opened 2023-04-03
Update AccessTokenHandler to use defaultazurecredential to fetch accesstoken

poadhika:
## Description
Update AccessTokenHandler to use defaultazurecredential to fetch accesstoken

## Related issues
Addresses [userstory #102164](https://microsofthealth.visualstudio.com/Health/_workitems/edit/102164).

## Testing
Did manual e2e testing. Also updated the unittests

Patch|Skip|Feature|Breaking


src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -1,29 +1,24 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
 using System.Threading;
 using System.Threading.Tasks;
-using EnsureThat;
-using Microsoft.Azure.Services.AppAuthentication;
```
wsugarman, line 9: Can you remove the associated NuGet package too?

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnectionBuilder.cs
```diff
@@ -16,7 +16,7 @@ public class ManagedIdentitySqlConnectionBuilder : ISqlConnectionBuilder
     private readonly ISqlConnectionStringProvider _sqlConnectionStringProvider;
     private readonly IAccessTokenHandler _accessTokenHandler;
     private readonly SqlRetryLogicBaseProvider _sqlRetryLogicBaseProvider;
-    private readonly string _azureResource = "https://database.windows.net/";
+    private readonly string _azureResourceScope = "https://database.windows.net/.default";
```
wsugarman, line 19: nit: should be `const`

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -1,29 +1,24 @@
...
 public class ManagedIdentityAccessTokenHandler : IAccessTokenHandler
 {
-    private readonly AzureServiceTokenProvider _azureServiceTokenProvider;
-
-    public ManagedIdentityAccessTokenHandler(AzureServiceTokenProvider azureServiceTokenProvider)
+    /// <inheritdoc />
+    public async Task<string> GetAccessTokenAsync(string resource, CancellationToken cancellationToken)
     {
-        EnsureArg.IsNotNull(azureServiceTokenProvider, nameof(azureServiceTokenProvider));
+        DefaultAzureCredential credential = new DefaultAzureCredential();
```
wsugarman, line 21: It looks like under the hood there is a new [`WorkloadIdentityCredential`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/WorkloadIdentityCredential.cs) class. I would say the options here are probably:

1. Configure this default credential to only enable both `ManagedIdentityCredential` and `WorkloadIdentityCredential` OR
2. Add a new flow for the SQL library for workload identity explicitly -- like how we have a managed identity path today

poadhika: To be clear, you want me to leave the existing ManagedIdentityAccessTokenHandler as is and create a new path that uses new handler that uses (default or managedidentity credential) only when workloadidentity is being used?

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -1,29 +1,27 @@
...
 
-        _azureServiceTokenProvider = azureServiceTokenProvider;
-    }
+        ManagedIdentityCredential credential = new ManagedIdentityCredential();
 
-    /// <inheritdoc />
-    public Task<string> GetAccessTokenAsync(string resource, CancellationToken cancellationToken)
-    {
-        return _azureServiceTokenProvider.GetAccessTokenAsync(resource, cancellationToken: cancellationToken);
+        var token = await credential.GetTokenAsync(new TokenRequestContext(new[] { resource }), CancellationToken.None).ConfigureAwait(false);
```
wsugarman, line 23: nit: you should forward the `CancellationToken` from the input

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
wsugarman: I still don't see a change to remove it from both the `csproj` file and the `Directory.packages.props`. Presumably this is `Microsoft.Azure.Services.AppAuthentication`?

wsugarman (changes requested):

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -1,29 +1,27 @@
...
-    public ManagedIdentityAccessTokenHandler(AzureServiceTokenProvider azureServiceTokenProvider)
+    /// <inheritdoc />
+    public async Task<string> GetAccessTokenAsync(string resource, CancellationToken cancellationToken)
     {
-        EnsureArg.IsNotNull(azureServiceTokenProvider, nameof(azureServiceTokenProvider));
+        EnsureArg.IsNotNullOrEmpty(resource, nameof(resource));
 
-        _azureServiceTokenProvider = azureServiceTokenProvider;
-    }
+        ManagedIdentityCredential credential = new ManagedIdentityCredential();
```
wsugarman, line 21: Do we need a hook for passing the `clientId`? Or is that unnecessary?

poadhika: That is not necessary. It reads from environment variables.

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -83,19 +82,6 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
...
         services.AddSingleton<IAccessTokenHandler, ManagedIdentityAccessTokenHandler>();
-        services.AddSingleton<AzureServiceTokenProvider>(p =>
-        {
-            SqlServerDataStoreConfiguration config = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value;
-
-            string tokenProviderConnectionString = null;
-
-            if (!string.IsNullOrEmpty(config.ManagedIdentityClientId))
-            {
-                tokenProviderConnectionString = $"RunAs=App;AppId={config.ManagedIdentityClientId}";
```
wsugarman, line 94: Should we now pass the client id as part of the handler?

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
poadhika: This code path gets executed only when one uses managed identity as their authentication. Also i verifed the authentication with both aad pod identity  and workload identity for our instances.

poadhika: But to be safe side updated to hook the clientid

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -82,19 +81,18 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
...
         {
             SqlServerDataStoreConfiguration config = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value;
 
-            string tokenProviderConnectionString = null;
+            string managedIdentityClientId = null;
 
             if (!string.IsNullOrEmpty(config.ManagedIdentityClientId))
             {
-                tokenProviderConnectionString = $"RunAs=App;AppId={config.ManagedIdentityClientId}";
+                managedIdentityClientId = config.ManagedIdentityClientId;
```
wsugarman, line 92: I think we can just pass `config.ManagedIdentityClientId` directly to the ctor

wsugarman: Or, simply update the ctor for `ManagedIdentityAccessTokenHandler` to use `IOptions<SqlServerDataStoreConfiguration>` to avoid this custom factory

src/Microsoft.Health.SqlServer/ManagedIdentityAccessTokenHandler.cs
```diff
@@ -1,29 +1,35 @@
...
     }
 
-    /// <inheritdoc />
-    public Task<string> GetAccessTokenAsync(string resource, CancellationToken cancellationToken)
+    public async Task<string> GetAccessTokenAsync(string resource, CancellationToken cancellationToken)
     {
-        return _azureServiceTokenProvider.GetAccessTokenAsync(resource, cancellationToken: cancellationToken);
+        EnsureArg.IsNotNullOrEmpty(resource, nameof(resource));
+
+        ManagedIdentityCredential credential = new ManagedIdentityCredential(_managedIdentityClientId);
```
wsugarman, line 29: Should we create this in the `ctor` and save it as a member instead?

poadhika: We can but this also works. Also curious what is the downside of doing it this way?

wsugarman (approved):

PR #733, opened 2023-05-22
Added workload identity access token handler

poadhika:
## Description
Added workload access token handler

## Related issues
Addresses [issue # 102674](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Medical%20Imaging/Stories/?workitem=102674)

## Testing
Did e2e testing with local cluster

Patch|Skip|Feature|Breaking


src/Microsoft.Health.SqlServer/WorkloadIdentityAcessTokenHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.Identity;
+using Microsoft.Health.SqlServer.Configs;
+
+namespace Microsoft.Health.SqlServer;
+public class WorkloadIdentityAcessTokenHandler : IAccessTokenHandler
```
wsugarman, line 13: Do we need a whole new handler for this? Could we instead use something like a [`ChainedTokenCredential`](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.chainedtokencredential?view=azure-dotnet) to combine both `ManagedIdentityCredential` and `WorkloadIdentityCredential`?

wsugarman (changes requested):

src/Microsoft.Health.SqlServer/WorkloadIdentityAcessTokenHandler.cs
poadhika: Yes that is possible. The reason why i added a new tokenhandler is because, i didnot wanted to change the original approach of using azureservicetokenprovider to fetch the new token. I have also added a new authentication type called worklaod identity. That way i am containing the affect the change i am making will have. For example, this change will  only take in effect if authentication is worklaodidentity. I am also planning to run few tests before on tip and low speed regions before enabling in all regions, Since we already have workload identity enabled in all regions, i found this approach will allow me to take control on switching back and forth between managed identity and workload identity.

wsugarman: That makes sense, and I see in the [KEDA codebase](https://github.com/kedacore/keda/blob/0a1f7a3c46cf083e3609942302b2521d9c1b7949/pkg/scalers/azure/azure_azidentity_chain.go#L24) they express a concern around identity mismatches between the Pod Identity and the Workload Identity, so perhaps you're right in that it's best to be explicit.

However, I think we can instead of creating `IAccessTokenHandler` classes, can we change the existing `ManagedIdentityAccessTokenHandler` class to be more like an `AzureCredentialTokenHandler`? Or we could perhaps even get rid of the interface altogether and pass an `AzureCredential` directly to the `ManagedIdentitySqlConnectionBuilder`? That'd probably be the simplest solution.

poadhika: Addressed the comments . Please take a look

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnectionBuilder.cs
```diff
@@ -14,25 +14,23 @@ namespace Microsoft.Health.SqlServer;
 public class ManagedIdentitySqlConnectionBuilder : ISqlConnectionBuilder
 {
     private readonly ISqlConnectionStringProvider _sqlConnectionStringProvider;
-    private readonly IAccessTokenHandler _accessTokenHandler;
     private readonly SqlRetryLogicBaseProvider _sqlRetryLogicBaseProvider;
-    private readonly string _azureResource = "https://database.windows.net/";
+    private readonly IAzureTokenCredentialProvider _azureTokenCredentialProvider;
 
     public ManagedIdentitySqlConnectionBuilder(
+        IAzureTokenCredentialProvider azureTokenCredentialProvider,
```
wsugarman, line 21: Do we need yet another abstraction, or can we simply rely on `AzureCredential` directly? Then in the registration method, we pass in the appropriate credential based on the type.

poadhika: I added it to support unit testing of ManagedIdentitySqlConnection.

wsugarman: But can't this connection builder just use `AzureCredential` instead of this abstraction? Like, we could just call this thing like `AzureCredentialSqlConnectionBuilder`

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnectionBuilder.cs
```diff
@@ -14,25 +15,24 @@ namespace Microsoft.Health.SqlServer;
 public class ManagedIdentitySqlConnectionBuilder : ISqlConnectionBuilder
```
wsugarman, line 15: We should probably rename this to like `CredentialedSqlConnectionBuilder`?

src/Microsoft.Health.SqlServer/ManagedIdentitySqlConnectionBuilder.cs
```diff
@@ -44,7 +44,13 @@ public async Task<SqlConnection> GetSqlConnectionAsync(string initialCatalog = n
             cancellationToken).ConfigureAwait(false);
 
         // set managed identity access token
-        sqlConnection.AccessToken = await _accessTokenHandler.GetAccessTokenAsync(_azureResource, cancellationToken).ConfigureAwait(false);
+        sqlConnection.AccessToken = await GetAccessTokenAsync(cancellationToken).ConfigureAwait(false);
         return sqlConnection;
     }
+
+    private async Task<string> GetAccessTokenAsync(CancellationToken cancellationToken = default)
```
wsugarman, line 51: nit: remove the default for a `private` member

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -70,32 +71,29 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
...
-
-            string tokenProviderConnectionString = null;
-
-            if (!string.IsNullOrEmpty(config.ManagedIdentityClientId))
+            p =>
             {
-                tokenProviderConnectionString = $"RunAs=App;AppId={config.ManagedIdentityClientId}";
-            }
+                var sqlServerDataStoreConfigOption = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>();
+                SqlServerDataStoreConfiguration config = sqlServerDataStoreConfigOption.Value;
```
wsugarman, line 77: nit:
```suggestion
                SqlServerDataStoreConfiguration config = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value;
```

src/Microsoft.Health.SqlServer.UnitTests/ManagedIdentitySqlConnectionTests.cs
```diff
@@ -18,47 +21,62 @@ public class ManagedIdentitySqlConnectionTests
...
+    [InlineData(SqlServerAuthenticationType.WorkloadIdentity)]
+    public async Task GivenDefaultConnectionType_WhenSqlConnectionToMasterRequested_MasterDatabaseIsSet(SqlServerAuthenticationType sqlServerAuthenticationType)
     {
+        InitializeTest(sqlServerAuthenticationType);
         SqlConnection sqlConnection = await _sqlConnectionFactory.GetSqlConnectionAsync(MasterDatabase).ConfigureAwait(false);
 
         Assert.Equal(MasterDatabase, sqlConnection.Database);
     }
+
+    private void InitializeTest(SqlServerAuthenticationType sqlServerAuthenticationType)
```
wsugarman, line 64: You can presumably move this back, right? As `ManagedIdentitySqlConnectionBuilder` doesn't actually read the `AuthenticationType` property

src/Microsoft.Health.SqlServer.UnitTests/ManagedIdentitySqlConnectionTests.cs
```diff
@@ -18,47 +21,62 @@ public class ManagedIdentitySqlConnectionTests
...
+
+        TokenCredential tokenCredential = Substitute.For<TokenCredential>();
+
+        AccessToken token = new AccessToken(TestAccessToken, DateTime.Now.AddHours(5));
+
+        tokenCredential.GetToken(new TokenRequestContext(new[] { _azureResource }), CancellationToken.None).Returns(token);
+
+        var sqlConfigOptions = Options.Create(sqlServerDataStoreConfiguration);
+        _sqlConnectionFactory = new ManagedIdentitySqlConnectionBuilder(
+            new DefaultSqlConnectionStringProvider(sqlConfigOptions), SqlConfigurableRetryFactory.CreateNoneRetryProvider(), tokenCredential);
```
wsugarman, line 80: Presumably the `DefaultSqlConnectionStringProvider` should be mocked via `NSubstitue` as well, right?

src/Microsoft.Health.SqlServer.UnitTests/ManagedIdentitySqlConnectionTests.cs
wsugarman: Then we can move it back to the ctor

src/Microsoft.Health.SqlServer.UnitTests/ManagedIdentitySqlConnectionTests.cs
poadhika: This value is still needed in tests that verifies the database name.

wsugarman (approved):

src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs
```diff
@@ -70,32 +70,26 @@ public static IServiceCollection AddSqlServerConnection(this IServiceCollection
...
-            }
+                SqlServerDataStoreConfiguration config = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value;
+                ISqlConnectionStringProvider sqlConnectionStringProvider = p.GetRequiredService<ISqlConnectionStringProvider>();
+                SqlRetryLogicBaseProvider sqlRetryLogic = p.GetRequiredService<SqlRetryLogicBaseProvider>();
 
-            return new AzureServiceTokenProvider(connectionString: tokenProviderConnectionString);
-        });
+                if (config.AuthenticationType == SqlServerAuthenticationType.ManagedIdentity)
+                {
+                    string managedIdentityClientId = p.GetRequiredService<IOptions<SqlServerDataStoreConfiguration>>().Value.ManagedIdentityClientId;
```
wsugarman, line 81: nit: you already fetched this service

PR #739, opened 2023-06-06
Schema Manager retry for Validating Version Compatibility

abdullah248:
## Description
Add retry when validating version compatibility as there are cases where a retry would succeed due to ttl in instance schema db.

Patch

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
@@ -63,34 +63,28 @@ public class SqlSchemaManager : ISchemaManager
             // since the Schema job polls and upserts at the specified interval in the service.
             await _baseSchemaRunner.EnsureInstanceSchemaRecordExistsAsync(token).ConfigureAwait(false);
 
-            int attemptCount = 1;
             int retryCountForHttpRequestException = 18;
 
             var availableVersions = await Policy.Handle<HttpRequestException>()
-            .WaitAndRetryAsync(
-                retryCount: retryCountForHttpRequestException,   // approx. 3 mins wait time for the service to responds to requests
-                sleepDurationProvider: (retryCount) => TimeSpan.FromSeconds(10),
-                onRetry: (exception, retryCount) =>
-                {
-                    _logger.LogError(exception, "Attempt {Attempt} of {MaxAttempts} to wait for the server to get started.", attemptCount++, retryCountForHttpRequestException);
-                })
-            .ExecuteAsync(token => GetAvailableSchema(token), token)
-            .ConfigureAwait(false);
+                .WaitAndRetryAsync(
+                    retryCount: retryCountForHttpRequestException,   // approx. 3 mins wait time for the service to responds to requests
+                    sleepDurationProvider: retryCount => TimeSpan.FromSeconds(10),
+                    onRetry: (exception, sleepDuration, retryCount, context) =>
+                        _logger.LogWarning(exception, "Attempt {Attempt} of {MaxAttempts} to wait for the server to get started.", retryCount, retryCountForHttpRequestException))
+                .ExecuteAsync(t => GetAvailableSchema(t), token)
+                .ConfigureAwait(false);
 
             // If the user hits apply command multiple times in a row, then the service schema job might not poll the updated available versions
             // so there are retries to give it a fair amount of time.
-            attemptCount = 1;
 
             availableVersions = await Policy.Handle<SchemaManagerException>()
-            .WaitAndRetryAsync(
-                retryCount: RetryAttempts,
-                sleepDurationProvider: (retryCount) => RetrySleepDuration,
-                onRetry: (exception, retryCount) =>
-                {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -119,7 +119,18 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
                 throw new SchemaManagerException(string.Format(CultureInfo.CurrentCulture, Resources.SpecifiedVersionNotAvailable, targetVersion));
             }
 
-            await ValidateVersionCompatibility(availableVersions[^1].Id, token).ConfigureAwait(false);
+            // TTL in instance schema table can cause this call to fail on first attempt so allow for retries
+            attemptCount = 1;
+            await Policy.Handle<SchemaManagerException>()
+            .WaitAndRetryAsync(
```
wsugarman, line 125: nit: indent the methods by 1

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -119,7 +119,18 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
                 throw new SchemaManagerException(string.Format(CultureInfo.CurrentCulture, Resources.SpecifiedVersionNotAvailable, targetVersion));
             }
 
-            await ValidateVersionCompatibility(availableVersions[^1].Id, token).ConfigureAwait(false);
+            // TTL in instance schema table can cause this call to fail on first attempt so allow for retries
+            attemptCount = 1;
```
wsugarman, line 123: The `onRetry` delegate is providing this information; you don't need to track it yourself

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -119,7 +119,18 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
...
+            await Policy.Handle<SchemaManagerException>()
+            .WaitAndRetryAsync(
+                retryCount: RetryAttempts,
+                sleepDurationProvider: (retryCount) => TimeSpan.FromSeconds(60),
```
wsugarman, line 127: nit you don't need `()` for single parameter delegates

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -119,7 +119,18 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
...
+                onRetry: (exception, retryCount) =>
+                {
+                    _logger.LogError(exception, "Attempt {Attempt} of {MaxAttempts} of validating version compatiblity", attemptCount++, RetryAttempts);
```
wsugarman, line 130: Should this be `Warning` instead?

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -119,7 +119,18 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
...
 
-            await ValidateVersionCompatibility(availableVersions[^1].Id, token).ConfigureAwait(false);
+            // TTL in instance schema table can cause this call to fail on first attempt so allow for retries
+            attemptCount = 1;
+            await Policy.Handle<SchemaManagerException>()
+            .WaitAndRetryAsync(
+                retryCount: RetryAttempts,
+                sleepDurationProvider: (retryCount) => TimeSpan.FromSeconds(60),
+                onRetry: (exception, retryCount) =>
+                {
```
wsugarman, line 129: nit: use expression over block (which I guess is a type of expression) for single line delegates

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -119,7 +119,18 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
...
+                    _logger.LogError(exception, "Attempt {Attempt} of {MaxAttempts} of validating version compatiblity", attemptCount++, RetryAttempts);
+                })
+            .ExecuteAsync(token => ValidateVersionCompatibility(availableVersions[^1].Id, token), token)
```
wsugarman, line 132: nit: rename the token to `t` in `ExcecuteAsync` to avoid the confusion

wsugarman (commented):
I see a number of my comments also pertain to the existing "retryable" code too

abdullah248:
Yeah I had reused the exisiting code. I updated all the existing ones and my changes as well. Also reason for previously they were using `attempt` to track number of retries is because they were using the wrong delegate which was actually providing the sleep duration, not the number of retries. I updated to use the right delegate which includes the retryCount. 

`WaitAndRetryAsync(int retryCount, Func<int, TimeSpan> sleepDurationProvider, Action<Exception, TimeSpan, int, Context> onRetry)`

PR #777, opened 2023-08-11
Add health check for availability of customer managed key

jnlycklama:
## Description
- Added "CustomerKeyValidationBackgroundService" - which will test the accessability of the key every min and the result to "ICustomerManagedKeyStatus". This test will get, wrap, and unwrap the key. All actions are required for CMK
- Added HealthCheckPublisher; it will run once every 10 seconds, executing all health checks and storing the result. /health/check endpoint will now read from this result instead of calling the registered health checks.
- ICustomerManagedKeyStatus is added to blob and SQL health checks so they can act based on the results
- If CMK is not enabled, the background service is still there, but it is a no-op and registers as healthy

## Related issues
Addresses [#104759] & [#104763].

## Testing
Ran code in dicom-server, tested out /health/check, behavior disabling/reenabling the key, and running without CMK configured

Patch


src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
```diff
@@ -37,13 +38,14 @@ public static void UseHealthChecksExtension(this IApplicationBuilder app, string
...
-            ResponseWriter = async (httpContext, healthReport) =>
+            ResponseWriter = async (httpContext, _) =>
             {
+                HealthReport healthReport = HealthCheckPublisher.Latest;
                 var response = JsonConvert.SerializeObject(
                     new
                     {
-                        overallStatus = healthReport.Status.ToString(),
-                        details = healthReport.Entries.Select(entry => new
+                        overallStatus = healthReport == null ? HealthStatus.Unhealthy.ToString() : healthReport?.Status.ToString(),
```
jnlycklama, line 47: This means /health/check will return Unhealthy until `HealthCheckPublisher.Latest` is populated. Currently the delay for `HealthCheckPublisher` to be run is 20 seconds to give time for the CMK health check to return a status (~15 seconds). So /health/check will return Unhealthy for the first 20-25 seconds.

An alternative is running all the health checks when we see `HealthCheckPublisher.Latest` is null. The only weird thing here would be when CMK is down - we would consistently report status as Degraded, but then when the service/conatiner restarts, the health check would momentarily report Healthy again before the CMK background service completes its first test.

src/Microsoft.Health.Core/Features/Health/HealthCheckPublisherExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public static IServiceCollection AddHealthCheckPublisher(this IServiceCollection services)
+    {
+        EnsureArg.IsNotNull(services, nameof(services));
+
+        services.Configure<HealthCheckPublisherOptions>(options =>
+        {
+            // 20 second start-up delay to give prerequisite health checks (ex: customer-managed keys) time to initialize
+            options.Delay = TimeSpan.FromSeconds(20);
+            options.Period = TimeSpan.FromSeconds(10);
```
jnlycklama, line 23: This means the HealthCheckPublisher will run all health checks every 10 seconds. Today we only cache health check data for 1 second, but its only refreshed when someone requests /health/check. With this, it will be run regardless, so 1s seems too frequent.

src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
jnlycklama: Talked with Will about this offline; decided to instead make the calls to the cached health report async, so /health/check will wait if it doesn't have all of the data. This means the initial call will take 10-15 seconds, but subsequent calls will be very quick

src/Microsoft.Health.Core/Features/Health/AsyncCache.cs
```diff
@@ -0,0 +1,38 @@
...
+        return _cachedData;
+    }
+
+    public void SetCachedData(T cachedData)
+    {
+        EnsureArg.IsNotNull(cachedData, nameof(cachedData));
+
+        _cachedData = cachedData;
+
+        _initializationTask.TrySetResult(true);
```
wsugarman, line 36: This actually returns a `bool` result that you would need to check, but furthermore I don't think a `TaskCompletionSource` can have its result updated as that brings it to a terminal state. So unfortunately I think you have to write the class differently

Ultimately we just need an async read-writer pattern where the publisher is a single writer for multiple middleware instances potentially reading the cached report. For example, you could use a `Task<HealthReport>` as the medium, where initially you have a `TaskCompletionSource<HealthReport>.Task` for readers to read, but over time you atomically swap out that task with new ones with updated reports with like `Task.FromResult(/* my health report */)`. You probably don't even need a new class for this. The publisher itself could implement both `IHealthCheckPublisher` as well as some read-only version of `IHealthCheckReportCache` (maybe just `IHealthReportCache`)?

src/Microsoft.Health.Core/Microsoft.Health.Core.csproj
```diff
@@ -1,12 +1,16 @@
...
     <Description>Common primitives and utilities used by Microsoft Health.</Description>
     <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Core" />
+    <PackageReference Include="Azure.Identity" />
     <PackageReference Include="Ensure.That" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" />
```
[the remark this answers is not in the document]
jnlycklama: IHealthCheckPublisher & HealthReport. But it looks like I don't need the Abstractions pkg anymore

src/Microsoft.Health.CustomerManagedKey/Client/EmptyKeyClient.cs
```diff
@@ -0,0 +1,13 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Azure.Security.KeyVault.Keys;
+
+namespace Microsoft.Health.CustomerManagedKey.Client;
+
+internal class EmptyKeyClient : KeyClient
```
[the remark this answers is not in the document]
jnlycklama: This actually wasn't included for test purposes - In `CustomerKeyClientFactory`, if I return `null` when CMK is not enabled, DI will complain because I'm trying to inject a null value for KeyClient. So I made this to be used for when CMK is not enabled. Not sure if theres a better way to do this instead

src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
```diff
@@ -57,4 +59,20 @@ public static void UseHealthChecksExtension(this IApplicationBuilder app, string
             },
         });
     }
+
+    /// <summary>
+    /// Use health checks (extension method). Register the response as json.
+    /// </summary>
+    /// <param name="app">Application builder instance.</param>
+    /// <param name="healthCheckPathString">Health check path string.</param>
+    public static void UseCachedHealthChecks(this IApplicationBuilder app, string healthCheckPathString)
```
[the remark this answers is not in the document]
jnlycklama: Since this is registering on `IApplicationBuilder`, and the publisher is on `IServiceCollection`, I thought it would be best to keep them separate

src/Microsoft.Health.Encryption/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            _customerManagedKeyHealth.SetCachedData(new CustomerKeyHealth
+            {
+                IsHealthy = true,
+                Description = null,
+                Reason = ExternalHealthReason.None,
+                Exception = null,
+            });
+        }
+        catch (Exception ex) when (ex is RequestFailedException || ex is CryptographicException || ex is InvalidOperationException || ex is NotSupportedException)
```
jnlycklama, line 66: I ended up keeping this logic in the background service so I was able to keep the unit tests on this behavior without involving the actual KeyClient calls

src/Microsoft.Health.CustomerManagedKey/Microsoft.Health.CustomerManagedKey.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
jnlycklama: Ended up just changing the name to Microsoft.Health.Encryption

src/Microsoft.Health.CustomerManagedKey/Client/EmptyKeyClient.cs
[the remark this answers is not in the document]
jnlycklama: Talked offline; ended up creating the KeyClient in the test itself, so we don't need to worry about creating a null KeyClient

src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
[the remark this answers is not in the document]
jnlycklama: Talked offline; added a check to ensure the publisher was registered

src/Microsoft.Health.Core/Features/Health/AsyncCache.cs
jnlycklama: Talked offline; ended up continuing with TaskCompletionSource for its simplicity and ability to await, but cleaned it up a little bit. Added CancellationToken, and only set result if the task has not been completed yet

src/Microsoft.Health.Core/Features/Health/AsyncData.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Core.Features.Health;
+
+/// <summary>
+/// Asynchronously waits for the data to be set before the data can be retrieved
+/// </summary>
+/// <typeparam name="T">The data type</typeparam>
+public class AsyncData<T> where T : class
```
jnlycklama, line 16: Renamed this to AsyncData to not conflict with https://github.com/microsoft/dicom-server/blob/0d916dd9f5d8c6b27c0b434dbb9ef1ac64283d0e/src/Microsoft.Health.Dicom.Core/Features/Common/AsyncCache.cs#L13

src/Microsoft.Health.Encryption/Health/KeyWrapUnwrapTestProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+    private readonly CustomerManagedKeyOptions _customerManagedKeyOptions;
+
+    public KeyWrapUnwrapTestProvider(
+        IExternalCredentialProvider credentialProvider,
+        IOptions<CustomerManagedKeyOptions> cmkOptions)
+    {
+        EnsureArg.IsNotNull(credentialProvider, nameof(credentialProvider));
+        EnsureArg.IsNotNull(cmkOptions, nameof(cmkOptions));
+
+        _customerManagedKeyOptions = cmkOptions.Value;
```
wsugarman, line 32:
```suggestion
        _customerManagedKeyOptions = EnsureArg.IsNotNull(cmkOptions?.Value, nameof(cmkOptions));
```

src/Microsoft.Health.Encryption/Health/KeyWrapUnwrapTestProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Core.Features.Identity;
+using Microsoft.Health.Encryption.Configs;
+
+namespace Microsoft.Health.Encryption.Health;
+
+internal class KeyWrapUnwrapTestProvider : IKeyTestProvider
+{
+    private readonly KeyClient _keyClient;
+
```
wsugarman, line 22: nit: no line between fields

src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
```diff
@@ -57,4 +61,25 @@ public static void UseHealthChecksExtension(this IApplicationBuilder app, string
             },
         });
     }
+
+    /// <summary>
+    /// Use health checks (extension method). Register the response as json.
+    /// </summary>
+    /// <param name="app">Application builder instance.</param>
+    /// <param name="healthCheckPathString">Health check path string.</param>
+    public static void UseCachedHealthChecks(this IApplicationBuilder app, PathString healthCheckPathString)
```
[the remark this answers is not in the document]
jnlycklama: I don't see why we wouldn't want to require it? Otherwise Map will map the middleware to empty path /

src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
```diff
@@ -57,4 +61,25 @@ public static void UseHealthChecksExtension(this IApplicationBuilder app, string
...
+    {
+        EnsureArg.IsNotNull(app, nameof(app));
+
+        // ensure IHealthCheckPublisher has been registered
+        app.ApplicationServices.GetRequiredService(typeof(IHealthCheckPublisher));
```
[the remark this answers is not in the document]
jnlycklama: Updated to check for HealthCheckPublisher

src/Microsoft.Health.Blob/Features/Health/BlobHealthCheck.cs
```diff
@@ -31,30 +36,46 @@ public class BlobHealthCheck : IHealthCheck
...
 
     public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
     {
         _logger.LogInformation("Performing health check.");
+
+        CustomerKeyHealth cmkStatus = await _customerKeyHealthCache.GetCachedData(cancellationToken).ConfigureAwait(false);
+        if (!cmkStatus.IsHealthy)
+        {
+            // if the customer-managed key is inaccessible, storage will also be inaccessible
+            return new HealthCheckResult(
```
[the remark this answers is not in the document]
jnlycklama: I feel like it makes sense to leave the `HealthCheckResult` in the `IHealthCheck`s. This way each health check can decide how they want to react when their dependency is unhealthy

src/Microsoft.Health.Blob/Microsoft.Health.Blob.csproj
```diff
@@ -8,13 +8,15 @@
...
     <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Azure.Security.KeyVault.Keys" />
     <PackageReference Include="Azure.Storage.Blobs" />
     <PackageReference Include="Azure.Storage.Common" />
     <PackageReference Include="Ensure.That" />
     <PackageReference Include="Microsoft.Extensions.Azure" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" />
```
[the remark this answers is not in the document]
jnlycklama: I dont think so, removed

src/Microsoft.Health.Api/Registration/ApplicationBuilderExtensions.cs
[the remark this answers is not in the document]
jnlycklama: Jk, updated to check for AsyncData<HealthReport> since that is what we actually need

src/Microsoft.Health.Blob/Features/Health/BlobHealthCheck.cs
```diff
@@ -31,30 +36,46 @@ public class BlobHealthCheck : IHealthCheck
...
-    /// <param name="testProvider">The test provider.</param>
+    /// <param name="testProvider">The blob test provider.</param>
+    /// <param name="customerKeyHealthCache">The status of the customer-managed key</param>
     /// <param name="logger">The logger.</param>
     public BlobHealthCheck(
         BlobServiceClient client,
         IOptionsSnapshot<BlobContainerConfiguration> namedBlobContainerConfigurationAccessor,
         string containerConfigurationName,
         IBlobClientTestProvider testProvider,
+        AsyncData<CustomerKeyHealth> customerKeyHealthCache,
```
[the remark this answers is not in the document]
jnlycklama: Looks like nobody else uses the BlobHealthCheck, but it is an issue for SqlHealthCheck. I added the registration for the CMK background service to the SQL registration. If CMK options are empty, it will skip keyvault check and always return healthy

src/Microsoft.Health.Blob/Features/Health/BlobHealthCheck.cs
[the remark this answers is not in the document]
jnlycklama: Then we can also use this data in the future in our exception handling, to determine whether or not the error is due to CMK issue

src/Microsoft.Health.Encryption/Extensions/CustomerKeyRegistrationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        return services;
+    }
+
+    public static IServiceCollection AddCustomerKeyHealthTest(this IServiceCollection services)
+    {
+        EnsureArg.IsNotNull(services, nameof(services));
+
+        services.AddOptions<CustomerManagedKeyOptions>();
+        services.TryAddSingleton<IExternalCredentialProvider, DefaultExternalCredentialProvider>();
+        services.TryAddSingleton<IKeyTestProvider, KeyWrapUnwrapTestProvider>();
```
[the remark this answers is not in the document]
jnlycklama: For `IExternalCredentialProvider` - yes, this should be override in paas to add in a non-default provider. 
For `IKeyTestProvider` - no. Updated to `AddSingleton`

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -19,17 +22,30 @@ public class SqlServerHealthCheck : IHealthCheck
 {
     private readonly ILogger<SqlServerHealthCheck> _logger;
     private readonly SqlConnectionWrapperFactory _sqlConnectionWrapperFactory;
+    private readonly AsyncData<CustomerKeyHealth> _customerKeyHealthCache;
 
-    public SqlServerHealthCheck(SqlConnectionWrapperFactory sqlConnectionWrapperFactory, ILogger<SqlServerHealthCheck> logger)
+    public SqlServerHealthCheck(SqlConnectionWrapperFactory sqlConnectionWrapperFactory, AsyncData<CustomerKeyHealth> customerKeyHealthCache, ILogger<SqlServerHealthCheck> logger)
```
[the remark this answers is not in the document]
jnlycklama: Responded in other comment

src/Microsoft.Health.Core/Features/Health/AsyncData.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public void SetCachedData(T cachedData)
+    {
+        EnsureArg.IsNotNull(cachedData, nameof(cachedData));
+
+        _cachedData = cachedData;
+
+        if (!_init.Task.IsCompleted)
+        {
+            _init.TrySetResult();
```
wsugarman, line 35: I am unsure about the thread-safety implications of this class's method. It may be best to try to use a `lock`

src/Microsoft.Health.Core/Features/Health/AsyncData.cs
wsugarman: Maybe `Cache<T>`? `ResultCache<T>`? `ValueCache<T>`? `SingletonCache<T>`?

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyHealth.cs
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Microsoft.Health.Core.Features.Health;
+
+namespace Microsoft.Health.Encryption.Customer.Health;
+
+public class CustomerKeyHealth
```
wsugarman, line 11: nit: some XML doc comments

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            try
+            {
+                await CheckHealth(stoppingToken).ConfigureAwait(false);
+                await Task.Delay(TimeSpan.FromSeconds(60), stoppingToken).ConfigureAwait(false);
```
wsugarman, line 43: Should this be configurable in `CustomerManagedKeyOptions`?

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+            catch (TaskCanceledException e)
```
wsugarman, line 45: nit: `OperationCanceledException` is more generic

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,77 @@
...
+            await _keyTestProvider.PerformTestAsync(cancellationToken).ConfigureAwait(false);
+
+            _customerManagedKeyHealth.SetCachedData(new CustomerKeyHealth
+            {
+                IsHealthy = true,
+                Reason = ExternalHealthReason.None,
+                Exception = null,
+            });
+        }
+        catch (Exception ex) when (ex is RequestFailedException || ex is CryptographicException || ex is InvalidOperationException || ex is NotSupportedException)
```
wsugarman, line 65: If this fails for any other reason, the entire background service will crash. Should we catch other exceptions and set the health appropriately? Or let the app crash?

src/Microsoft.Health.Encryption/Customer/Health/IKeyTestProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Encryption.Customer.Health;
+
+internal interface IKeyTestProvider
+{
+    Task PerformTestAsync(CancellationToken cancellationToken = default);
```
wsugarman, line 13: Maybe something like `AssertHealthAsync` as "Asserts" typically throw exceptions when they fail, like this method

src/Microsoft.Health.Encryption/Microsoft.Health.Encryption.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Encryption utilities, including those for customer-managed keys, used by Microsoft Health.</Description>
+    <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Security.KeyVault.Keys" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" />
```
wsugarman, line 10: Double-checking again, is this really necessary?

src/Microsoft.Health.Encryption.UnitTests/Microsoft.Health.Encryption.UnitTests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
```
wsugarman, line 6: Is this necessary?

src/Microsoft.Health.Encryption/Customer/Extensions/CustomerKeyRegistrationExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        services.AddSingleton<IKeyTestProvider, KeyWrapUnwrapTestProvider>();
+
+        if (configure != null)
+        {
+            services.Configure(configure);
+        }
+        else
+        {
+            // allows enabling the background service when CustomerManagedKeyOptions has not been provided
+            services.Configure<CustomerManagedKeyOptions>(c => c = new CustomerManagedKeyOptions());
```
wsugarman, line 36: This should be unnecessary for `IOptions<T>` as configuration is optional

src/Microsoft.Health.Core/Features/Health/AsyncData.cs
jnlycklama: I don't know if that is really necessary. It looks like TaskCompletionSource promises thread safety (https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource-1?view=net-7.0#thread-safety), and if I look under the hood at `TrySetResult` it does seem like its handling it for us.
(https://github.com/microsoft/referencesource/blob/master/mscorlib/system/threading/Tasks/TaskCompletionSource.cs#L296)
```
        public bool TrySetResult(TResult result)
        {
            bool rval = m_task.TrySetResult(result);
            if (!rval && !m_task.IsCompleted) SpinUntilCompleted();
            return rval;
        }
```
Where `SpinUntilCompleted` is there to wait in case another thread is trying to set the result

src/Microsoft.Health.Core/Microsoft.Health.Core.csproj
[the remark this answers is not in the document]
jnlycklama: Ah looks like we do need it, but not for the class I said. HealthCheckPublisherOptions is not in abstractions https://github.com/dotnet/aspnetcore/blob/f71e2c5598abf6ac639be5936bd77650b879e026/src/HealthChecks/HealthChecks/src/HealthCheckPublisherOptions.cs#L11

src/Microsoft.Health.Core/Features/Health/AsyncData.cs
jnlycklama: Renamed to `ValueCache<T>`

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
jnlycklama: Added new property `KeyValidationPeriodInSeconds`, but left 60 seconds as the default

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
jnlycklama: Yeah good call, the app shouldn't crash on this failing. I added a catch all afterwards, which will log the exception and reset the state to healthy. This will avoid flagging CMK misconfigurations when we are failing for other reasons like bugs / outages

src/Microsoft.Health.Encryption/Microsoft.Health.Encryption.csproj
jnlycklama: Removed from Encryption project

src/Microsoft.Health.Encryption.UnitTests/Microsoft.Health.Encryption.UnitTests.csproj
jnlycklama: No, but when I was creating this initially I was having the hardest time getting VS to recognize it as a test project. After I removed these it would switch to a Library... but now it seems to work

src/Microsoft.Health.Encryption/Customer/Extensions/CustomerKeyRegistrationExtensions.cs
jnlycklama: Removed this, but had to add null checks where its used (in the key test provider and cmk background validation service)

src/Microsoft.Health.Core/Features/Health/AsyncData.cs
wsugarman: Ah, I missed this remark in the docs! Thanks for checking!

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheckMiddleware.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System.Threading.Tasks;
+using System;
+using EnsureThat;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System.Collections.Generic;
+using Microsoft.Health.Core.Features.Health;
+using System.Linq;
+using System.Collections.Immutable;
```
wsugarman, line 14: These using statements are out-of-order

src/Microsoft.Health.Core/Features/Health/ValueCache.cs
```diff
@@ -0,0 +1,39 @@
...
+/// Caches a single value of type <typeparamref name="T"/>.
+/// Asynchronously waits for the data to be set before the data can be retrieved.
+/// </summary>
+/// <typeparam name="T">The data type</typeparam>
+public class ValueCache<T> where T : class
+{
+    private volatile T _cachedData;
+    private readonly TaskCompletionSource _init = new TaskCompletionSource();
+
+    public async Task<T> GetCachedData(CancellationToken cancellationToken = default)
```
wsugarman, line 22: nit: Maybe just `GetAsync` and `Set`

src/Microsoft.Health.Core/Features/HealthPublisher/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,30 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Microsoft.Health.Core.Features.Health;
+
+internal class HealthCheckPublisher : IHealthCheckPublisher
```
wsugarman, line 13: Maybe something like `InMemoryHealthCheckPublisher`?

src/Microsoft.Health.Core/Features/HealthPublisher/HealthCheckPublisherExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.Health.Core.Features.Health;
+
+public static class HealthCheckPublisherExtensions
+{
+    public static IServiceCollection AddHealthCheckPublisher(this IServiceCollection services, Action<HealthCheckPublisherOptions> configure = null)
+    {
+        EnsureArg.IsNotNull(configure, nameof(configure));
+
+        services.AddSingleton<IHealthCheckPublisher, HealthCheckPublisher>();
```
wsugarman, line 19: I would use `TryAddEnumerable` here instead as the idea is 1+ health check publishers can be added

src/Microsoft.Health.Core/Microsoft.Health.Core.csproj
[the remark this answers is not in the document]
wsugarman: Boo!

src/Microsoft.Health.Encryption/Customer/Configs/CustomerManagedKeyOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public const string CustomerManagedKey = "CustomerManagedKey";
+
+    public Uri KeyVaultUri { get; set; }
+
+    public string KeyVersion { get; set; }
+
+    public string KeyName { get; set; }
+
+    public int KeyValidationPeriodInSeconds { get; set; }
```
wsugarman, line 20: nit: use a `TimeSpan` and use name `ValidationPeriod`

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyHealth.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.Health.Encryption.Customer.Health;
+
+/// <summary>
+/// The health of the encryption key provided by the customer
+/// </summary>
+public class CustomerKeyHealth
+{
+    /// <summary>
+    /// Specifies if the resource is healthy or not
```
wsugarman, line 17: You don't need to fix this, but typically in the base class libraries (BCL) they use the phrases `Gets [or sets]...` for their XML doc

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Core.Features.Health;
+using Microsoft.Health.Encryption.Customer.Configs;
+
+namespace Microsoft.Health.Encryption.Customer.Health;
+
+internal class CustomerKeyValidationBackgroundService : BackgroundService
+{
+    private const string AccessLostMessage = "Access to the customer-managed key has been lost";
+    private const int DefaultKeyValidationPeriodInSeconds = 60;
```
wsugarman, line 23: Specify the default in the options class

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private readonly TimeSpan _keyValidationPeriod;
+
+    public CustomerKeyValidationBackgroundService(
+        IKeyTestProvider keyTestProvider,
+        ValueCache<CustomerKeyHealth> customerManagedKeyHealth,
+        IOptions<CustomerManagedKeyOptions> customerManagedKeyOptions,
+        ILogger<CustomerKeyValidationBackgroundService> logger)
+    {
+        EnsureArg.IsNotNull(customerManagedKeyOptions, nameof(customerManagedKeyOptions));
```
wsugarman, line 37: Assert that `customerManagedKeyOptions?.Value` is not `null`

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "This background service failing should not crash the application.")]
+    internal async Task CheckHealth(CancellationToken cancellationToken)
+    {
+        try
+        {
+            await _keyTestProvider.AssertHealthAsync(cancellationToken).ConfigureAwait(false);
+            SetHealthy();
+        }
+        catch (Exception ex) when (ex is RequestFailedException || ex is CryptographicException || ex is InvalidOperationException || ex is NotSupportedException)
```
wsugarman, line 72: nit: use pattern-matching!
```suggestion
        catch (Exception ex) when (ex is RequestFailedException or CryptographicException or InvalidOperationException or NotSupportedException)
```

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheckMiddleware.cs
```diff
@@ -0,0 +1,72 @@
...
+        HealthReport latestReport = await _healthCheckReportCache.GetCachedData(httpContext.RequestAborted).ConfigureAwait(false);
+
+        if (!DefaultStatusCodesMapping.TryGetValue(latestReport.Status, out var statusCode))
+        {
+            var message = $"No status code mapping found for {nameof(HealthStatus)} value: {latestReport.Status}.";
+            throw new InvalidOperationException(message);
+        }
+
+        httpContext.Response.StatusCode = statusCode;
+        await httpContext.Response.WriteAsJsonAsync(FormatReport(latestReport)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jnlycklama: I will probably leave this as is then to keep consistent validation with the existing APIs

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
jnlycklama: Ah I had assumed that if a consumer didn't specify these cmk options that `customerManagedKeyOptions?.Value` would be null, and we had to handle that case. But looks like it just creates an empty object

src/Microsoft.Health.Core/Features/HealthPublisher/HealthCheckPublisher.cs
jnlycklama: I changed it to `HealthCheckCachePublisher` to be more similar to the other naming - `ValueCache`, `CachedHealthCheckMiddleware`, etc.

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -69,9 +60,8 @@ internal async Task CheckHealth(CancellationToken cancellationToken)
             await _keyTestProvider.AssertHealthAsync(cancellationToken).ConfigureAwait(false);
             SetHealthy();
         }
-        catch (Exception ex) when (ex is RequestFailedException || ex is CryptographicException || ex is InvalidOperationException || ex is NotSupportedException)
+        catch (Exception ex) when (ex is CustomerKeyInaccessibleException)
```
wsugarman, line 63: nit:
```suggestion
        catch (CustomerKeyInaccessibleException ex)
```

jnlycklama: Ah forgot to update the formatting :(

wsugarman (dismissed):

PR #792, opened 2023-08-31
Refactor SQL Connection Authentication

wsugarman:
## Description
This Pull Request refactors how authentication works for SQL connections. Historically, we have modified the `SqlConnection` objects by having separate abstractions for the generation of SQL connection strings and their corresponding connections. However, the SQL client SDK [supports managed identity and other authentication schemes natively](https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver16) and as such we can rely on those instead of our own logic that relies on deprecated libraries. In this PR, I remove the need for the `ISqlConnectionStringProvider` interface and add its logic into the existing `ISqlConnectionBuilder` interface.

Going forward, instead of relying on `AuthenticationType`, we can specify managed identity authentication directly in the connection string.

## Related issues
[AB#106461](https://microsofthealth.visualstudio.com/Health/_workitems/edit/106461/)

## Testing
New and existing tests. I have already tested this ad-hoc in K8s using connection strings.

+semver: major


src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapperFactory.cs
```diff
@@ -37,6 +37,8 @@ public class SqlConnectionWrapperFactory
         _sqlRetryLogicBaseProvider = sqlRetryLogicBaseProvider;
     }
 
+    public string DefaultDatabase => _sqlConnectionBuilder.DefaultDatabase;
```
poadhika, line 40: What is the default database here?

poadhika:
Did we verify that the change is working with dicomservices which has Workloadidentity enabled?

src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
```diff
@@ -52,6 +53,7 @@ public class SqlServerDataStoreConfiguration
     /// <summary>
     /// If set, the client id of the managed identity to use when connecting to SQL, if AuthenticationType == ManagedIdentity.
     /// </summary>
+    [Obsolete("Update ConnectionString to supply User ID instead.")]
```
poadhika, line 56: User ID or managed identity?

wsugarman: `User ID` as that is the field in the SQL connection string

wsugarman:
> Did we verify that the change is working with dicomservices which has Workloadidentity enabled?

Yup! I have an instance EUAP I've been experimenting with

src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
wsugarman: The connection strings should look like these: https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver16#using-active-directory-managed-identity-authentication

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -268,6 +264,8 @@ public static void ValidateDatabaseName(string databaseName)
         {
             throw new InvalidOperationException("The initial catalog in the connection string cannot be a system database");
         }
+
+        return databaseName;
```
bcarthic, line 268: What is the purpose of returning the database that is passed as an argument?

src/Microsoft.Health.SqlServer/Features/Client/SqlConnectionWrapper.cs
```diff
@@ -60,8 +60,7 @@ internal async Task InitializeAsync(string initialCatalog = null, CancellationTo
         }
         else
         {
-            SqlConnection = await _sqlConnectionBuilder.GetSqlConnectionAsync(initialCatalog, _sqlServerDataStoreConfiguration.MaxPoolSize, cancellationToken: cancellationToken).ConfigureAwait(false);
-            SqlConnection.RetryLogicProvider = _sqlRetryLogicBaseProvider;
```
bcarthic, line 64: I see you have removed the retry logic provider and is it just sufficient to have `services.AddSqlRetryLogicProvider();` this in the registration? Did we test out if in case of timeout, we retry the specified number of times?

tools/SchemaManager/Program.cs
```diff
@@ -63,7 +63,7 @@ private static ServiceProvider BuildServiceProvider(string[] args)
         // TODO: this won't work in OSS if the AuthenticationType is set to ManagedIdentity
         services.AddSingleton<ISqlConnectionBuilder, DefaultSqlConnectionBuilder>();
 
-        services.TryAddSingleton<SqlRetryLogicBaseProvider>(p =>
+        services.TryAddSingleton(p =>
```
bcarthic, line 66: Don't we need to add the retryProvider before DefaultConnectionBuilder? We do that in `SqlServerBaseRegistrationExtensions`

tools/SchemaManager/Program.cs
```diff
@@ -63,7 +63,7 @@ private static ServiceProvider BuildServiceProvider(string[] args)
         // TODO: this won't work in OSS if the AuthenticationType is set to ManagedIdentity
```
bcarthic, line 63: Is this True now?

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
wsugarman: - To assign the result after valuation. Just for convenience

tools/SchemaManager/Program.cs
wsugarman: Order doesn't matter for registration of different services. It only makes for the same service type

tools/SchemaManager/Program.cs
wsugarman: I think this just refers to the the use of connection strings in OSS? It would work today too if we used managed identities in app services.

rbans96 (dismissed):

bcarthic (approved):

bcarthic (approved):

PR #820, opened 2023-09-28
Catch errors where sql db is inacessible

jnlycklama:
## Description
Scenario missed in health state:
- cmk is enabled
- customer misconfigures cmk, and leaves it misconfigured for 30 mins
- we report Degraded status because we cannot access the key
- customer fixes their cmk configuration, but does not reprovision the service in order to revalidate the key on the SQL DB
- we are able to access the key, so health checks will pass the CMK check, try to connect to the SQL DB, and fail

Adding a new state here so we can display more details to the customer about why their service might be in this state, and what they can do to fix it (reprovision/update)

## Related issues
Addresses [issue #].

## Testing

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
@@ -1,4 +1,4 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -90,6 +90,12 @@ public async Task ExecuteAsync(SchemaInformation schemaInformation, string insta
             {
                 // this could happen during schema initialization until base schema is not executed so can be ignored
             }
+            // Error: "Can not connect to the database in its current state". This error can be for various DB states (recovering, inacessible) but we assume that our DB will only hit this for Inaccessible state
+            catch (SqlException ex) when (ex.ErrorCode == 40925)
+            {
+                // this can happen when a user has misconfigured CMK for > 30 min, which results in the DB having a status of "Inacessible".
+                _logger.LogInformation(ex, "The SQL database cannot be accessed in its current state.");
```
wsugarman, line 97: Warning?

wsugarman (approved):

src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
jnlycklama: I've been logging all of the CMK errors as Information so far - I know DRIs sometimes will check the warnings summary in the Exceptions dashboard and I don't want this to be anything people would need to investigate, since it is not our fault

PR #845, opened 2023-10-19
Create abstract class to define StorageHealthChecks that are dependent on CMK health

jnlycklama:
## Description
Changes:
- StorageHealthCheck. Defines how health checks should interact with CMK health. First - check cmk health cache. If its unhealthy, return Degraded. If healthy, continue with storage health check. Catch any exceptions related to CMK, including SqlException for the state being inaccessible. This state can stay Inaccessible even after the key is accessible.
- Update background service to not log on cancellation token thrown
- Key test provider returns status instead of throwing an exception

## Testing

Patch|Skip|Feature|Breaking


src/Microsoft.Health.Encryption/Customer/Health/IDataStoreStateTestProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Encryption.Customer.Health;
+
+internal interface IInaccessibleStateTestProvider
+{
+    Task AssertHealthAsync(CancellationToken cancellationToken = default);
```
wsugarman, line 13: I am generally not a fan of exception-based flow control. One of the reasons being that it is quite slow. Can this interface return something instead?

src/Microsoft.Health.SqlServer/Features/Health/SQLInaccessibleStateTestProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+    public async Task AssertHealthAsync(CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            using SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken).ConfigureAwait(false);
+        }
+        // Error: Can not connect to the database in its current state. This error can be for various DB states (recovering, inacessible) but we assume that our DB will only hit this for Inaccessible state
+        catch (SqlException ex) when (ex.ErrorCode == 40925)
+        {
+            _logger.LogInformation(ex, InaccessibleMessage);
```
wsugarman, line 40: Maybe `Warning`?

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -21,6 +22,8 @@ namespace Microsoft.Health.SqlServer.Features.Health;
 /// </summary>
 public class SqlServerHealthCheck : IHealthCheck
 {
+    // This health check will not pass if the cached health is failing for any of these reasons
+    private readonly IEnumerable<HealthStatusReason> _dependentHealthStatusReasons = new List<HealthStatusReason> { HealthStatusReason.CustomerManagedKeyAccessLost, HealthStatusReason.DataStoreStateDegraded };
```
wsugarman, line 26: nit: same comment as blob

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -42,7 +45,9 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         _logger.LogInformation($"Starting {nameof(SqlServerHealthCheck)}.");
 
         CustomerKeyHealth cmkStatus = await _customerKeyHealthCache.GetAsync(cancellationToken).ConfigureAwait(false);
-        if (!cmkStatus.IsHealthy)
+
+        if (!cmkStatus.IsHealthy &&
```
wsugarman, line 49: nit: one line

wsugarman, line 49: Something still smells funny about this. Essentially every health check that relies on CMK makes the same check with some modifications. I would say either:
1. We use some sort of abstract class for these health checks that check CMK to perform the preconditions
2. Manage the relationships between health checks outside of their implementations ala some sort of decorator pattern

wsugarman (changes requested):

src/Microsoft.Health.SqlServer/Features/Health/SQLInaccessibleStateTestProvider.cs
jnlycklama: If we log this as a warning, it will show up in our dashboards and cause noise for DRIs. Since this is not something people should be investigating, I think its still ok to leave as Information. It will still show up in the Exceptions table since the exception is added

src/Microsoft.Health.Blob/Features/Health/BlobHealthCheck.cs
```diff
@@ -23,6 +24,8 @@ namespace Microsoft.Health.Blob.Features.Health;
 /// </summary>
 public class BlobHealthCheck : IHealthCheck
 {
+    // This health check will not pass if the cached health is failing for any of these reasons
+    private readonly IEnumerable<HealthStatusReason> _dependentHealthStatusReasons = new List<HealthStatusReason> { HealthStatusReason.CustomerManagedKeyAccessLost };
```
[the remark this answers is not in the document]
jnlycklama: Resolving b/c outdated

src/Microsoft.Health.Blob/Features/Health/BlobHealthCheck.cs
```diff
@@ -67,7 +70,8 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         _logger.LogInformation("Performing health check.");
 
         CustomerKeyHealth cmkStatus = await _customerKeyHealthCache.GetAsync(cancellationToken).ConfigureAwait(false);
-        if (!cmkStatus.IsHealthy)
+        if (!cmkStatus.IsHealthy &&
+            _dependentHealthStatusReasons.Contains(cmkStatus.Reason))
```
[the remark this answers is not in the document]
jnlycklama: Resolving b/c outdated

src/Microsoft.Health.Encryption/Customer/Health/IDataStoreStateTestProvider.cs
jnlycklama: Hm ok - so the reason I threw the exception was because I wanted to exception to be passed down to the CMK Health, which would then get passed to the health checks / health state. 

I updated this to instead just return CustomerKeyHealth, which contains the exception. Also ended up merging these into a single interface since they are doing / returning the same thing

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
jnlycklama: Resolving b/c outdated

jnlycklama: I agree it was a little weird - refactored it a bit to make "dependent" classes. `KeyAccessDependent` and `KeyAccessAndDataStoreStateDependent`, which define which reasons that they care about. The health checks can inherit this (and possibly other things in the future? could help with exception handling) and use it to decide whether or not to continue after getting the CMK health.
LMK what you think

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,66 @@
...
+                cmkStatus.Exception,
+                new Dictionary<string, object> { { "Reason", cmkStatus.Reason } });
+        }
+
+        try
+        {
+            await CheckStorageHealthAsync(cancellationToken).ConfigureAwait(false);
+            return HealthCheckResult.Healthy("Successfully connected.");
+        }
+        catch (Exception ex) when (CMKAccessLostExceptionFilter(ex))
```
jnlycklama, line 49: Based on the flakiness in the e2e tests I created, it looks like the underlying storgae (storage acct or sql) can take a little bit to come online after the key is restored. So we should catch errors here as well to not crash

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,73 @@
...
+                sqlEx,
+                new Dictionary<string, object> { { "Reason", HealthStatusReason.DataStoreStateDegraded } });
+        }
+    }
+
+    public abstract string DegradedDescription { get; }
+
+    public abstract Func<Exception, bool> CMKAccessLostExceptionFilter { get; }
+
+    public abstract Task CheckStorageHealthAsync(CancellationToken cancellationToken);
```
wsugarman, line 72: Should we allow instead for callers to return `Task<HealthCheckResult>` so that they have the flexibility to return unhealthy results or change the properties of healthy results?

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,73 @@
...
+        catch (Exception ex) when (CMKAccessLostExceptionFilter(ex))
+        {
+            return new HealthCheckResult(
+                HealthStatus.Degraded,
+                DegradedDescription,
+                ex,
+                new Dictionary<string, object> { { "Reason", HealthStatusReason.CustomerManagedKeyAccessLost } });
+        }
+        // Error: "Can not connect to the database in its current state". This error can be for various DB states (recovering, inacessible) but we assume that our DB will only hit this for Inaccessible state
+        catch (SqlException sqlEx) when (sqlEx.ErrorCode == 40925)
```
wsugarman, line 58: Is it possible to push this into the SQL health check instead? E.g. does the blob storage health check care about this? Ideally we'd remove any dependency on SQL here (if possible).

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,73 @@
...
+                HealthStatus.Degraded,
+                DegradedDescription,
+                sqlEx,
+                new Dictionary<string, object> { { "Reason", HealthStatusReason.DataStoreStateDegraded } });
+        }
+    }
+
+    public abstract string DegradedDescription { get; }
+
+    public abstract Func<Exception, bool> CMKAccessLostExceptionFilter { get; }
```
wsugarman, line 70: If we really need to keep this functionality based on my other comments, let's use a method instead. `abstract` function properties are pretty atypical.

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            return new HealthCheckResult(
+                HealthStatus.Degraded,
+                DegradedDescription,
+                sqlEx,
+                new Dictionary<string, object> { { "Reason", HealthStatusReason.DataStoreStateDegraded } });
+        }
+    }
+
+    public abstract string DegradedDescription { get; }
```
wsugarman, line 68: Is this worth making configurable? If so, should it be `virtual` instead?

wsugarman (changes requested):

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
jnlycklama: Good call, updated to virtual

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
jnlycklama: Makes sense. I dont know why in my head it seemed like this had to be a property lol

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
jnlycklama: I had it here initially because I wanted it to go after the CMKAccessLost exception catch, since that should take priority. But the exception will only ever be one thing lol so it works the same

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyDependent.cs
```diff
@@ -3,12 +3,9 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
-using System.Threading;
-using System.Threading.Tasks;
-
 namespace Microsoft.Health.Encryption.Customer.Health;
 
-internal interface IKeyTestProvider
+public abstract class CustomerKeyDependent
```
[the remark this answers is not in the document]
jnlycklama: File is deleted

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyConstants.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Azure;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Health.Encryption.Customer.Health;
+
+public static class CustomerKeyConstants
```
wsugarman, line 12: Can't you define these as methods in their corresponding health check?

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyConstants.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public static Func<Exception, bool> StorageAccountExceptionFilter => ex => ex is RequestFailedException rfe && rfe.ErrorCode == "KeyVaultEncryptionKeyNotFound";
+
+    /// <summary>
+    /// Filter on error codes for azure key vault https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-31000-to-41399?view=sql-server-ver16
+    /// Intentionally leaving out code 40925 when the DB is Inaccessible so that can be reported using a different HealthStatusReason
+    /// </summary>
+    public static Func<Exception, bool> SQLExceptionFilter => ex => ex is SqlException sqlException && (sqlException.ErrorCode == 40981 || sqlException.ErrorCode == 33183 || sqlException.ErrorCode == 33184);
```
wsugarman, line 20: You can use the fancy patter-matching syntax here. I think the syntax would be:

```suggestion
    public static Func<Exception, bool> SQLExceptionFilter => ex => ex is SqlException sql && (sql.ErrorCode is 40981 or 33183 or 33184);
```

src/Microsoft.Health.Encryption/Microsoft.Health.Encryption.csproj
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Security.KeyVault.Keys" />
+    <PackageReference Include="Microsoft.Data.SqlClient" />
```
wsugarman, line 10: I think if you move the "constants" you can remove this

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,62 @@
...
+                HealthStatus.Degraded,
+                DegradedDescription,
+                ex,
+                new Dictionary<string, object> { { "Reason", HealthStatusReason.CustomerManagedKeyAccessLost } });
+        }
+    }
+
+    public virtual string DegradedDescription => "The health of the store has degraded.";
+
+    public abstract bool CMKAccessLostExceptionFilter(Exception ex);
```
wsugarman, line 59: nit: use a verb for methods.

```suggestion
public abstract bool IsCMKAccessLost(Exception ex);
```

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyConstants.cs
jnlycklama: DurableTaskHubHealthCheck will need the storage filter. and not sure if any checks for FHIR will need the SQL filter, but thought since storage is here, it made sense for both of them to be here

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyConstants.cs
jnlycklama: cool! updated

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyConstants.cs
jnlycklama: Discussed offline, created new abstract classes to cover it

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            // if the customer-managed key is inaccessible, storage will also be inaccessible
+            return new HealthCheckResult(
+                HealthStatus.Degraded,
+                DegradedDescription,
+                cmkStatus.Exception,
+                new Dictionary<string, object> { { "Reason", cmkStatus.Reason } });
+        }
+
+        try
```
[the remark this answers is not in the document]
wsugarman: E.g.

Perhaps in the core project?

```csharp
public static class HealthStatusReasonExtensions
{
    // This method can create a HealthCheckResult per reason
    // We can have a test that enumerates through the values (Enum.GetValues<HealthStatusReason>()) to ensure they work
    public static HealthCheckResult ToHealthCheckResult(
        this HealthStatusReason reason,
        Exception exception = null,
        string description = null)
    { /* ... */ }
}
```

src/Microsoft.Health.Encryption/Customer/Health/CustomerKeyValidationBackgroundService.cs
```diff
@@ -45,9 +46,9 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                 await CheckHealth(stoppingToken).ConfigureAwait(false);
                 await Task.Delay(_customerManagedKeyOptions.KeyValidationPeriod, stoppingToken).ConfigureAwait(false);
             }
-            catch (OperationCanceledException e)
+            catch (OperationCanceledException) when (stoppingToken.IsCancellationRequested)
             {
-                _logger.LogInformation(e, $"{nameof(CustomerKeyValidationBackgroundService)} cancelled");
+                break;
```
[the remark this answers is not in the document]
jnlycklama: I'll blame SchemaJobWorker & my habit of copying for this lol https://github.com/microsoft/healthcare-shared-components/blob/d9dadcbfb2b1e89aa1c719dbf0944e882f2fd330/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs#L112

src/Microsoft.Health.Encryption/Customer/Health/KeyWrapUnwrapTestProvider.cs
```diff
@@ -45,11 +46,11 @@ internal class KeyWrapUnwrapTestProvider : IKeyTestProvider
         }
     }
 
-    public async Task AssertHealthAsync(CancellationToken cancellationToken = default)
+    public async Task<CustomerKeyHealth> AssertHealthAsync(CancellationToken cancellationToken = default)
     {
         if (_keyClient == null)
             // customer-managed key is not enabled
-            return;
+            return new CustomerKeyHealth();
```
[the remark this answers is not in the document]
jnlycklama: I had thought about something like this but I didn't want to create an Extensions class just for this

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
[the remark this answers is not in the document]
jnlycklama: I think I get what you're saying but that extension/utility is confusing to me for some reason. I added the try/catch to the SQLHealthCheck/AzureStorageHealthCheck instead

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -62,16 +49,25 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false);
 
             _logger.LogInformation("Successfully connected to SQL database.");
+
             return HealthCheckResult.Healthy("Successfully connected.");
         }
-        // Error: Can not connect to the database in its current state. This error can be for various DB states (recovering, inacessible) but we assume that our DB will only hit this for Inaccessible state
-        catch (SqlException ex) when (ex.ErrorCode == 40925)
+        // Filter on error codes for azure key vault https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-31000-to-41399?view=sql-server-ver16
+        catch (Exception ex) when (ex is SqlException sqlException && (sqlException.ErrorCode is 40981 or 33183 or 33184))
```
wsugarman, line 56: Perhaps just combine the `catch` blocks and change the reason based on the error code

src/Microsoft.Health.Encryption/Customer/Health/AzureStorageHealthCheck.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+    }
+
+    public override async Task<HealthCheckResult> CheckStorageHealthAsync(CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await CheckAzureStorageHealthAsync(cancellationToken).ConfigureAwait(false);
+        }
+        catch (Exception ex) when (ex is RequestFailedException rfe && rfe.ErrorCode == "KeyVaultEncryptionKeyNotFound")
```
wsugarman, line 30: nit: just catch `RequestFailedException`

src/Microsoft.Health.Encryption/Customer/Health/StorageHealthCheck.cs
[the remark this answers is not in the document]
wsugarman: Sure, and if we really want something like this, we can add it later

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -62,17 +49,25 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false);
 
             _logger.LogInformation("Successfully connected to SQL database.");
+
             return HealthCheckResult.Healthy("Successfully connected.");
         }
-        // Error: Can not connect to the database in its current state. This error can be for various DB states (recovering, inacessible) but we assume that our DB will only hit this for Inaccessible state
-        catch (SqlException ex) when (ex.ErrorCode == 40925)
+        // Filter on error codes for azure key vault https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-31000-to-41399?view=sql-server-ver16
+        catch (Exception ex) when (ex is SqlException sqlException && (sqlException.ErrorCode is 40981 or 33183 or 33184 or 40925))
```
wsugarman, line 56: nit: just do something like:
```suggestion
        catch (SqlException e) when (e.ErrorCode is 40981 or 33183 or 33184 or 40925)
```

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -62,17 +49,25 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
         {
-            // DB is status in Inaccessible because the encryption key was inacessible for > 30 mins. User must reprovision or we need to revalidate key on SQL DB. 
+            HealthStatusReason reason = HealthStatusReason.CustomerManagedKeyAccessLost;
```
wsugarman, line 58: nit:
```suggestion
            HealthStatusReason reason = sqlException.ErrorCode is 40925 
                ? HealthStatusReason.DataStoreStateDegraded
                : HealthStatusReason.CustomerManagedKeyAccessLost
```

wsugarman (approved):

PR #847, opened 2023-10-20
Downgrade IdentityModel to 6.33 to support dependent packages

bcarthic:
## Description
Downgrade IdentityModel to 6.34 to support dependent packages

## Related issues
Addresses [issue #].

## Testing

Patch|Skip|Feature|Breaking

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <HighEntropyVA>true</HighEntropyVA>
-    <IdentityModelTokenPackageVersion>6.34.0</IdentityModelTokenPackageVersion>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -11,7 +11,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <HighEntropyVA>true</HighEntropyVA>
-    <IdentityModelTokenPackageVersion>7.0.3</IdentityModelTokenPackageVersion>
+    <IdentityModelTokenPackageVersion>6.34.0</IdentityModelTokenPackageVersion>
```
wsugarman, line 14: This version is unlisted in NuGet.org. Are you sure we want this version?

wsugarman (changes requested):

Directory.Build.props
bcarthic: Using the previous released version.

wsugarman (approved):

PR #876, opened 2023-11-14
[Health Check] Reason as string

fhibf:
Modifying the Health Check logic to return the failure reason as string, instead of an integer. 

The behavior I want to avoid is the following one, where the reason is returned as a number.

image


jnlycklama:
In `HealthCheckMetricPublisherTests` we are checking that we can correctly parse the reason that is being emitted by the healthy checks. Could you update this to use the string version as well & make sure the tests pass? https://github.com/microsoft/healthcare-shared-components/blob/98a206472b84338326e7ad79f8f5abef9dc3ad2c/src/Microsoft.Health.Core.UnitTests/Features/HealthPublisher/HealthCheckMetricPublisherTests.cs#L138

jnlycklama (approved):

PR #882, opened 2023-11-16
Add .NET 8 Target

wsugarman:
## Description
- Add `net8.0` target in addition to `net6.0` and `net7.0`
- Address new analyzer concerns
  - This includes the use of the new [`CompositeFormat`](https://learn.microsoft.com/en-us/dotnet/standard/base-types/composite-formatting) type. To more easily integrate these cached format strings, I have added a new type called `FormatResources` per project that contains cached format strings in .NET 8
- Remove usage of the `Clock` class and replace its usage with [`TimeProvider`](https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0) in .NET 8 and beyond

## Related issues
N/A

## Testing
Existing tests

+semver: feature


src/Microsoft.Health.SqlServer/Features/Schema/Manager/SchemaClient.cs
```diff
@@ -49,7 +49,7 @@ public async Task<string> GetScriptAsync(int version, CancellationToken cancella
         }
         else
         {
-            throw new SchemaManagerException(string.Format(CultureInfo.CurrentCulture, Resources.ScriptNotFound, response.StatusCode));
+            throw new SchemaManagerException(Resources.ScriptNotFound);
```
wsugarman, line 52: Some of these resources were not actually format strings and so these `string.Format` calls did not do anything

src/Microsoft.Health.SqlServer/Features/Schema/Manager/SqlSchemaManager.cs
```diff
@@ -65,7 +65,7 @@ public virtual async Task ApplySchema(MutuallyExclusiveType type, bool force = f
 
             int retryCountForHttpRequestException = 18;
 
-            var availableVersions = await Policy.Handle<HttpRequestException>()
+            _ = await Policy.Handle<HttpRequestException>()
```
wsugarman, line 68: It doesn't look like we actually needed to save this value as it was immediately re-assigned

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -38,7 +38,7 @@ public class AuditEventTypeMapping : IAuditEventTypeMapping, IHostedService
 {
     private readonly IActionDescriptorCollectionProvider _actionDescriptorCollectionProvider;
 
-    private IReadOnlyDictionary<(string ControllerName, string ActionName), Attribute> _attributeDictionary;
```
bcarthic, line 41: I see, in some place you use ImmutableDictionary compared to Dictionary. It is because that was a static property?

src/Microsoft.Health.Client/Authentication/OAuth2ClientCertificateCredentialProvider.cs
```diff
@@ -76,6 +76,7 @@ protected override async Task<string> BearerTokenFunction(CancellationToken canc
 
     internal static string GenerateClientAssertion(string clientId, X509Certificate2 x509Certificate2, Uri tokenUri)
     {
+        // TODO: Migrate to using JsonWebToken
```
bcarthic, line 79: Is this in this PR or for later?

src/Microsoft.Health.Core/Configs/AuditConfiguration.cs
```diff
@@ -12,29 +12,14 @@ public class AuditConfiguration
...
-        {
-            if (string.IsNullOrEmpty(value))
-            {
-                throw new InvalidDefinitionException(Resources.CustomHeaderPrefixCannotBeEmpty);
-            }
-
-            _customAuditHeaderPrefix = value;
-        }
+        get => _customAuditHeaderPrefix;
+        set => _customAuditHeaderPrefix = !string.IsNullOrEmpty(value) ? value : throw new InvalidDefinitionException(Resources.CustomHeaderPrefixCannotBeEmpty);
```
bcarthic, line 23: Why didnt we use CompositeResources here?

test/Microsoft.Health.SqlServer.Web/Microsoft.Health.SqlServer.Web.csproj
```diff
@@ -31,12 +36,12 @@
     <EmbeddedResource Include="Features\Schema\Migrations\3.sql" />
     <GenerateFilesInputs Include="Features\Schema\Migrations\2.sql" />
     <GenerateFilesInputs Include="Features\Schema\Migrations\3.sql" />
-    <Generated Include="Features\Schema\Model\VLatest.Generated.cs">
+    <Generated Include="Features\Schema\Model\VLatest.Generated.$(TargetFramework).cs">
```
bcarthic, line 39: Was there any specific need to this way? Did something broke with having older framework file?

Directory.Packages.props
```diff
@@ -71,10 +73,11 @@
...
+    <PackageVersion Include="xunit" Version="2.6.2" />
     <PackageVersion Include="xunit.assert" Version="2.6.2" />
     <PackageVersion Include="Xunit.DependencyInjection" Version="8.9.0" />
     <PackageVersion Include="xunit.extensibility.core" Version="2.6.2" />
-    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.3" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.4" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+
```
bcarthic, line 82: nit: was the extra line intentional.

src/Microsoft.Health.Api/Features/Audit/AuditEventTypeMapping.cs
wsugarman: It was because it is a `private` member and we don't necessarily care about its immutability as much, as only the instance has access. But there could be an argument for making it an `ImmutableDictionary<TKey, TValue>`

src/Microsoft.Health.Client/Authentication/OAuth2ClientCertificateCredentialProvider.cs
wsugarman: This is for later and anyone that looks at it. The `JsonWebToken` type seems to be newer and more performant

test/Microsoft.Health.SqlServer.Web/Microsoft.Health.SqlServer.Web.csproj
wsugarman: This is to prevent issues with concurrent builds for multiple build targets. In truth, this code should be generated using source generators instead.

Directory.Packages.props
wsugarman: Yeah, I like adding these as our other project files have them

bcarthic (approved):

PR #894, opened 2023-12-11
Add utility for performance measurement, tracking the duration of scoped code block.

ms-teli:
## Description
Add utility for performance measurement, tracking the duration of scoped code block.

## Testing
Unit Test was added.


src/Microsoft.Health.Utility/Microsoft.Health.Utility.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
dustinburson, line 1: Code itself looks good and I like the update to use a delegate to make the function more generic.  My question is do we need a new library or can this be part of an existing one like https://github.com/microsoft/healthcare-shared-components/tree/main/src/Microsoft.Health.Core?  It seems like a reasonable fit there.  If we keep this library, what is the guidance for when new code would go here vs other libraries?

dustinburson (approved):

src/Microsoft.Health.Utility/ITimed.cs
```diff
@@ -0,0 +1,15 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Utility;
+
+public interface ITimed : IDisposable
```
brendankowitz, line 10: Seems pretty similar to: 

https://github.com/microsoft/healthcare-shared-components/blob/09135f1a5717707d7d6137e8c5dbe35777dcb9b4/src/Microsoft.Health.Core/ActionTimer.cs#L13-L40

src/Microsoft.Health.Utility/Microsoft.Health.Utility.csproj
brendankowitz: > The utility tool bag can be consumed by other packages too with clear meaning

I'd ague the opposite 😄 by definition, a general purpose utility library's actual purpose can't be well defined

src/Microsoft.Health.Utility/ITimed.cs
ms-teli: Thanks for sharing. I am taking a look at it.

src/Microsoft.Health.Utility/Microsoft.Health.Utility.csproj
ms-teli: > > The utility tool bag can be consumed by other packages too with clear meaning
> 
> I'd argue the opposite 😄 by definition, a general purpose utility library's actual purpose can't be well defined

That sounds true. :) The other option I can think of is to rename it utility.measurements.

src/Microsoft.Health.Utility/ITimed.cs
ms-teli: I preferred the abstracted version since this class depends on the logger, in our case, it could be a telemetry client.

dustinburson (approved):

PR #896, opened 2023-12-15
Check for null lockHandle before using

feordin:
## Description
Update the check for the lockHandle being NULL before the using in SchemaInitializer.  If the lockHandle is null, and we pass in a NULL value to the using () clause, this can result in a Null Ref exception when .Net disposes of the resources from the using.

## Related issues
Addresses [issue AB#113443].

## Testing

Patch|Skip|Feature|Breaking


brendankowitz (approved):
:shipit:

wsugarman (approved):

PR #902, opened 2024-02-13
[SQL Health Check] Handle HTTP exceptions

fhibf:
SQL Health Check - Handle HTTP exception

## Related issues
- FHIR Health Checks can fail due AAD misconfiguration. And those failures can happen when one of our internal components try to acquire the connection string to SQL. 

## Testing
Added new unit test to ensure errors in SQL Health Checks are handled properly.


src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -52,18 +54,31 @@ public override async Task<HealthCheckResult> CheckStorageHealthAsync(Cancellati
 
             return HealthCheckResult.Healthy("Successfully connected.");
         }
-        catch (SqlException e) when (e.IsCMKError())
+        catch (HttpRequestException httpe) when (IsInvalidAccess(httpe))
```
jnlycklama, line 57: nit: spelling `http`

jnlycklama (commented):
Is there also some way to check if the access is controlled by the customer? Maybe we can pass in a config or something that will have a flag if the SQL connection is customer controlled?

I'm worried about this filtering out health check failures that are caused by actual bugs. For example, in Dicom SQL is completely internal so there's no way a customer could mess anything up. If we get a Forbidden or Unauthorized exception, we would want to throw an exception here because that means something to do with our token is messed up.

src/Microsoft.Health.SqlServer/Features/Health/SqlServerHealthCheck.cs
```diff
@@ -52,18 +54,31 @@ public override async Task<HealthCheckResult> CheckStorageHealthAsync(Cancellati
...
             return new HealthCheckResult(
                 HealthStatus.Degraded,
                 DegradedDescription,
-                e,
+                sqlEx,
                 new Dictionary<string, object> { { "Reason", reason.ToString() } });
         }
     }
+
+    private static bool IsInvalidAccess(HttpRequestException exception)
```
jnlycklama, line 82: Should we put this in a common location? So its not duplicated across here + `SqlServerSchemaDataStore`

jnlycklama, line 82: Adding a comment here to specify the scenario in which this could happen would be helpful. In the future we might see this and not remember or remove it

PR #914, opened 2024-03-13
Replace ADO Pipelines with GitHub Actions

wsugarman:
## Description
Move pipelines entirely to Github Actions. Includes a new PR and CI workflow. As part of the CI workflow:
- Packages are pushed to the existing public microsofthealthoss NuGet feed
- Symbols are pushed to the symbol server
- A GitHub release is created

## Testing
Ad-hoc


.github/workflows/packages-ci.yml
```diff
@@ -0,0 +1,68 @@
+name: Packages PR
```
PTaladay, line 1: Copy paste issue? Should be "Packages CI" correct?

wsugarman: Ah yes! Good catch!

.github/actions/dotnet-publish/action.yml
```diff
@@ -0,0 +1,59 @@
...
+      dotnet pack "./Microsoft.Health.sln" \
+        -c "${{ inputs.buildConfiguration }}" \
+        "-p:ContinuousIntegrationBuild=true" \
+        "-p:PackageVersion=${{ inputs.packageVersion }}" \
+        --no-build \
+        -o "${{ inputs.packageFolder }}"
+
+  - name: dotnet nuget push
+    shell: bash
+    working-directory: './packages'
```
wsugarman, line 40: Oops, have to change this working directory

.github/actions/github-release/action.yml
```diff
@@ -0,0 +1,32 @@
...
+    description: The title for the GitHub release
+    required: true
+
+runs:
+  using: composite
+  steps:
+  - name: Release
+    shell: bash
+    run: |
+      gh release create "${{ inputs.tag }}" \
```
wsugarman, line 23: Unsure if I need to tag the commit first. I'd imagine I have to... so I'll look how to best do it through our automation. I've done similar things before like [this](https://github.com/wsugarman/durabletask-azurestorage-scaler/blob/main/.github/actions/github-release/scripts/CreateRelease.mjs).

wsugarman: The `gh release create` command will automatically create a tag if necessary. I've passed the commit SHA as well to ensure it correctly tags the proper code. See https://cli.github.com/manual/gh_release_create

bcarthic (approved):

PR #932, opened 2024-03-28
Fixing InvalidOperationException from SchemaInitializer.

v-iyamauchi:
## Description
Fixing the exception below that occurs on SchemaInitalizer.InitializeAsync.

System.InvalidOperationException: Nullable object must have a value.

## Related issues
Addresses [issue #118108].

[Bug 118108](https://microsofthealth.visualstudio.com/Health/_workitems/edit/118108): IHost.StartAsync fails for SchemaManagerService due to exceptions from SchemaInitializer.

## Testing
Tested in the Fhir validation pipeline for health-paas.

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
@@ -169,7 +169,7 @@ public sealed class SchemaInitializer : IHostedService
 
         // Ensure to publish the Schema notifications even when schema is up-to date and Schema Initializer is called again (like restarting FHIR server will call this again)
         // There is a dependency on this notification in FHIR server to enable some background jobs
-        if (!schemaUpgradedNotificationSent)
+        if (!schemaUpgradedNotificationSent && _schemaInformation.Current.HasValue)
         {
             await _mediator.NotifySchemaUpgradedAsync((int)_schemaInformation.Current, false).ConfigureAwait(false);
         }


SergeyGaluzo (approved):
:shipit:

apurvabhaleMS (approved):

src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -169,7 +169,7 @@ await using (lockHandle.ConfigureAwait(false))
 
         // Ensure to publish the Schema notifications even when schema is up-to date and Schema Initializer is called again (like restarting FHIR server will call this again)
         // There is a dependency on this notification in FHIR server to enable some background jobs
-        if (!schemaUpgradedNotificationSent)
+        if (!schemaUpgradedNotificationSent && _schemaInformation.Current.HasValue)
```
brendankowitz, line 172: Question, is this is to allow logging when the application has failed to initialize? Not firing this message will cause other components to never initialize also

v-iyamauchi: I noticed this exception after I changed the AllowDatabaseCreation flag to false in ApplyPaasCommand (health-paas) that is preventing the generic host from starting up successfully. Without successful host startup, an activity listener in Otel tracing won't start listening to any activity causing schema apply operation logs (in FhirLog table) without the operation Id (empty operationId in AI table and env_dt_traceId in Otel table). 

Regardless of the above, the fix is needed to avoid the exception from casting a nullable object to int when the object is null.

System.InvalidOperationException: Nullable object must have a value.
   at Microsoft.Health.SqlServer.Features.Schema.SchemaInitializer.InitializeAsync(IServiceScope scope, Boolean forceIncrementalSchemaUpgrade, CancellationToken cancellationToken)
   at Microsoft.Health.SqlServer.Features.Schema.SchemaInitializer.InitializeAsync(Boolean forceIncrementalSchemaUpgrade, CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Health.Cloud.ServicePlatform.ServiceFabric.WorkerStatelessService`2.RunAsync(CancellationToken cancellationToken)

brendankowitz: Sounds like there are two critical problems, you can't log, but you also have no database schema. Are we attempting to start the service in a faulted state for the purpose of logging?

I thought the Model Initialization also relies on this? https://github.com/microsoft/fhir-server/blob/ae6128c7462298fe577656424fe24d35117736b9/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SchemaUpgradedHandler.cs#L14 however, this might not matter if the service is in an errored state anyway.

v-iyamauchi: My knowledge of the schema initialization/upgrade is limited, but here is what I observed.

1. SchemaInitializer implements IHostedService and it gets started during the generic host startup.
2. ApplyPaasCommand in health-paas was setting AllowDatabaseCreation flag to true that was causing a login failure to the [master database](https://github.com/microsoft/healthcare-shared-components/blob/ac6aee7cb8511e6e9fba760afdf2be768b1a579c/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs#L217C64-L217C65) resulting in a failure in SchemaInitializer initialization process.
3. After this [commit ](https://microsofthealth.visualstudio.com/Health/_git/health-paas/commit/6b8ab0f8ea5e7daa66899b9fa665b2a994143d3f?refName=refs%2Fheads%2Fmain)for setting AllowDatabaseCreation to false in ApplyPaasCommand, SchemaInitializer initialization process is still failing due to [invalid casting of a nullable object to integer](https://github.com/microsoft/healthcare-shared-components/blob/ac6aee7cb8511e6e9fba760afdf2be768b1a579c/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs#L174C55-L174C86) (because it is null).

This change will fix the invalid casting (throwing an invalid operation exception) causing the generic host startup (preventing an activity listener in Otel tracing from listening any activity). Please let me know if you have any more questions.

brendankowitz: This class feels like it is important for the initialization of FHIR Server which is why I'm asking. I see some [logging below](https://github.com/microsoft/healthcare-shared-components/blob/ba63e07555d6a2e504f2add4abf441c5f1d23e54/src/Microsoft.Health.SqlServer/Features/Schema/SchemaInitializer.cs#L197) that instruments that the schema is not found. Should be ok to merge.

v-iyamauchi: Thank you, @brendankowitz. Will merge the change.

PR #1077, opened 2024-11-04
Add Health Check Cache Option to Configure Which Results are Cached

wsugarman:
## Description
Add new options `MinimumCachedHealthStatus` to configure which results are cached

## Related issues
[AB#131890](https://microsofthealth.visualstudio.com/Health/_workitems/edit/131890)

## Testing
New unit tests

Feature

### DIFF
--- a/src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
+++ b/src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
@@ -23,8 +23,8 @@ internal sealed class CachedHealthCheck : IHealthCheck, IDisposable
     private readonly ILogger _logger;
     private readonly SemaphoreSlim _semaphore;
 
-    // By default, the times are DateTimeOffset.MinValue such that they are always considered invalid
-    private CachedHealthCheckResult _cache = new CachedHealthCheckResult();
+    // By default, the times are DateTimeOffset.MinValue such that they are always considered expired
+    private CachedHealthCheckResult _cache = new CachedHealthCheckResult { Result = new HealthCheckResult(HealthStatus.Unhealthy) };
 
 #if NET8_0_OR_GREATER
     private readonly TimeProvider _timeProvider;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
```diff
@@ -139,7 +139,12 @@ private bool HasExpired(DateTimeOffset timestamp)
         => timestamp >= _cache.ExpireTime;
 
     private bool IsUpToDate(DateTimeOffset timestamp)
-        => timestamp < _cache.StaleTime;
+    {
+        // While we will store results whose status is < MinimumCachedHealthStatus, it will effectively
+        // have no caching as IsUpToDate will return false
+        CachedHealthCheckResult currentCache = _cache;
+        return timestamp < currentCache.StaleTime && currentCache.Result.Status >= _options.MinimumCachedHealthStatus;
```
bcarthic, line 146: So if its unhealthy/degraded we always want to evaluate? I guess for most of the services this won't be an issue, but we have few services who are always unhealthy we may do lot of calls and can be throttled

bcarthic (approved):

src/Microsoft.Health.Api/Features/HealthChecks/CachedHealthCheck.cs
KnicKnic, line 146: Are these values being accessed in a thread safe manner

wsugarman: This is definitely true, and something we'll have to account for using this setting.

wsugarman: Yes! I am first saving the cache by assigning it to a local variable. The `_cache` variable itself is replaced atomically using `Interlocked.CompareExchange`

PR #1149, opened 2025-03-10
Add app name to the connection string

SergeyGaluzo:
Adds ability to set application name on the connection string

### DIFF
--- a/src/Microsoft.Health.SqlServer/DefaultSqlConnectionBuilder.cs
+++ b/src/Microsoft.Health.SqlServer/DefaultSqlConnectionBuilder.cs
@@ -82,6 +82,23 @@ public class DefaultSqlConnectionBuilder : ISqlConnectionBuilder
         return new SqlConnection(builder.ToString()) { RetryLogicProvider = _retryProvider };
     }
 
+    [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "Caller must dispose result.")]
+    public async Task<SqlConnection> GetSqlConnectionAsync(bool isReadOnly, string applicationName, CancellationToken cancellationToken)
+    {
+        SqlConnectionStringBuilder builder = await GetConnectionStringBuilderAsync(null, null).ConfigureAwait(false);
+        if (isReadOnly)
+        {
+            builder.ApplicationIntent = ApplicationIntent.ReadOnly;
+        }
+
+        if (!string.IsNullOrEmpty(applicationName))
+        {
+            builder.ApplicationName = applicationName;
+        }
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.SqlServer/DefaultSqlConnectionBuilder.cs
```diff
@@ -82,6 +82,23 @@ public async Task<SqlConnection> GetReadOnlySqlConnectionAsync(string initialCat
         return new SqlConnection(builder.ToString()) { RetryLogicProvider = _retryProvider };
     }
 
+    [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "Caller must dispose result.")]
+    public async Task<SqlConnection> GetSqlConnectionAsync(bool isReadOnly, string applicationName, CancellationToken cancellationToken)
```
SergeyGaluzo, line 86: This method will be used in FHIR for read only connections and connections required to have application name set.

src/Microsoft.Health.SqlServer/DefaultSqlConnectionBuilder.cs
```diff
@@ -82,6 +82,23 @@ public async Task<SqlConnection> GetReadOnlySqlConnectionAsync(string initialCat
...
+    {
+        SqlConnectionStringBuilder builder = await GetConnectionStringBuilderAsync(null, null).ConfigureAwait(false);
```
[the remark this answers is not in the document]
SergeyGaluzo: fixed

src/Microsoft.Health.SqlServer/DefaultSqlConnectionBuilder.cs
```diff
@@ -82,6 +82,23 @@ public async Task<SqlConnection> GetReadOnlySqlConnectionAsync(string initialCat
         return new SqlConnection(builder.ToString()) { RetryLogicProvider = _retryProvider };
     }
 
+    [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "Caller must dispose result.")]
+    public async Task<SqlConnection> GetSqlConnectionAsync(bool isReadOnly, string applicationName)
```
LTA-Thinking, line 86: Will these connections never need the initialCatalog and maxPoolSize settings the other options allow for?

SergeyGaluzo: I don't see any case when we would need master database or more than 100 connections for FHIR server.

LTA-Thinking (approved):

PR #1157, opened 2025-03-29
Create New Methods for Creating SQL Connections with More Flexibility

wsugarman:
## Description
Adds new method overloads that allow users to specify delegates on the connection string builder. The alternative would be creating more and more overloads for every connection property.

Feature

### DIFF
--- a/src/Microsoft.Health.SqlServer/ISqlConnectionBuilder.cs
+++ b/src/Microsoft.Health.SqlServer/ISqlConnectionBuilder.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.SqlServer/ISqlConnectionBuilder.cs
```diff
@@ -20,6 +21,25 @@ public interface ISqlConnectionBuilder
     /// <value>The default initial catalog if specified; otherwise <see langword="null"/>.</value>
     string DefaultDatabase { get; }
 
+    /// <summary>
+    /// Gets a SQL connection as additionally configured by the caller.
+    /// </summary>
+    /// <param name="configure">An optional delegate for configuring the connection.</param>
+    /// <returns>The SQL connection.</returns>
+    SqlConnection CreateConnection(Action<SqlConnectionStringBuilder> configure = null);
```
wsugarman, line 29: I eventually want to deprecate the old methods

PR #1384, opened 2026-04-21
Handle read-only database (SQL 3906) in SchemaJobWorker for geo-replicated secondary

bcarthic:
## Problem

In geo-redundancy deployments, secondary K8s clusters run DICOM pods that connect to a **read-only SQL replica** (via SQL Failover Group). Each pod runs `SchemaJobWorker` as a background service that polls every N seconds and performs two write operations:

1. **`UpsertInstanceSchemaInformationAsync`** — registers the pod's schema version in the `InstanceSchema` table
2. **`DeleteExpiredInstanceSchemaAsync`** — cleans up stale rows from the same table

Both are **writes** that fail on the read-only secondary with **SQL error 3906** ("Failed to update database because the database is read-only").

### Impact
- The generic `catch (Exception)` catches this and logs it as an **error every poll cycle**, creating persistent error-level log noise on all secondary pods
- `schemaInformation.Current` is **never set** because the upsert throws before returning the version number — secondary pods don't know their schema version, which can cause downstream issues
- This is a **steady-state problem** whenever geo-redundancy is enabled — not just during failover

## Root Cause

`SchemaJobWorker` has no awareness of read-only databases. It unconditionally attempts writes on every poll cycle regardless of whether the database is writable.

### 1. `SqlErrorCodes.cs`
Added `ReadOnlyDatabase = 3906` constant for the SQL Server "database is read-only" error.

### 2. `SchemaJobWorker.cs`
- Wrapped `UpsertInstanceSchemaInformationAsync` in an inner try/catch for `SqlErrorCodes.ReadOnlyDatabase`
- On the read-only path: calls `GetCurrentVersionAsync` (a `SELECT` query — safe on read-only DBs) to resolve the current schema version
- Filters to only `completed` schema versions (ignores `started`/`failed`)
- Skips `DeleteExpiredInstanceSchemaAsync` on the read-only path (the primary handles cleanup and it replicates over)
- Logs at `Information` level instead of `Error`

### Why this is safe for everyone
- SQL error 3906 **only fires on read-only databases**. Normal (primary) databases never hit this code path
- The catch uses a `when` filter — it only matches this exact error number. All other `SqlException`s still go through existing catch blocks unchanged
- All 4 existing unit tests continue to pass with no changes

## Tests

3 new unit tests in `SchemaJobWorkerTests.cs`:

| Test | Validates |
|---|---|
| `GivenReadOnlyDatabase_WhenUpsertThrows3906_SchemaVersionIsReadFromCurrentVersions` | `schemaInformation.Current` is correctly set from `GetCurrentVersionAsync` fallback |
| `GivenReadOnlyDatabase_WhenUpsertThrows3906_DeleteExpiredInstanceSchemaIsSkipped` | `DeleteExpiredInstanceSchemaAsync` is never called on read-only path |
| `GivenReadOnlyDatabase_WhenUpsertThrows3906_OnlyCompletedVersionsAreConsidered` | Only `completed` versions are used; `started`/`failed` are ignored |

**All 7 tests pass** (4 existing + 3 new).

## Files Changed

| File | Change |
|---|---|
| `src/Microsoft.Health.SqlServer/Features/Storage/SqlErrorCodes.cs` | Add `ReadOnlyDatabase = 3906` |
| `src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs` | Catch SQL 3906, read version via `GetCurrentVersionAsync`, skip delete |
| `src/Microsoft.Health.SqlServer.UnitTests/Features/Schema/SchemaJobWorkerTests.cs` | 3 new unit tests |

[AB#190055](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/190055)

### DIFF
--- a/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
+++ b/src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
@@ -68,7 +69,27 @@ public class SchemaJobWorker
                 ISchemaDataStore schemaDataStore = scope.ServiceProvider.GetRequiredService<ISchemaDataStore>();
 
                 int? previous = schemaInformation.Current;
-                schemaInformation.Current = await schemaDataStore.UpsertInstanceSchemaInformationAsync(instanceName, schemaInformation, cancellationToken).ConfigureAwait(false);
+                bool isReadOnlyDatabase = false;
+
+                try
+                {
+                    schemaInformation.Current = await schemaDataStore.UpsertInstanceSchemaInformationAsync(instanceName, schemaInformation, cancellationToken).ConfigureAwait(false);
+                }
+                catch (SqlException ex) when (ex.Number == SqlErrorCodes.ReadOnlyDatabase)
+                {
+                    // On a geo-replicated secondary, the database is read-only. The InstanceSchema
+                    // table is replicated from primary, so the secondary does not need to register.
+                    // Read the current schema version instead.
+                    isReadOnlyDatabase = true;
+                    _logger.LogInformation("Database is read-only (geo-replication secondary). Skipping instance schema upsert.");
+
+                    var versions = await schemaDataStore.GetCurrentVersionAsync(cancellationToken).ConfigureAwait(false);
+                    var completedVersions = versions.Where(v => v.Status == SchemaVersionStatus.completed).ToList();
+                    if (completedVersions.Count > 0)
+                    {
+                        schemaInformation.Current = completedVersions.Max(v => v.Id);
+                    }
+                }
 
                 // If there was a change in the schema version and this isn't the base schema
                 if (schemaInformation.Current != previous && schemaInformation.Current > 0)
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -68,7 +69,27 @@ public async Task ExecuteAsync(SchemaInformation schemaInformation, string insta
...
+                }
+                catch (SqlException ex) when (ex.Number == SqlErrorCodes.ReadOnlyDatabase)
+                {
+                    // On a geo-replicated secondary, the database is read-only. The InstanceSchema
+                    // table is replicated from primary, so the secondary does not need to register.
+                    // Read the current schema version instead.
+                    isReadOnlyDatabase = true;
+                    _logger.LogInformation("Database is read-only (geo-replication secondary). Skipping instance schema upsert.");
+
+                    var versions = await schemaDataStore.GetCurrentVersionAsync(cancellationToken).ConfigureAwait(false);
```
jnlycklama, line 86: Will there be a problem if the dicom pod is not compatible with the version read in the SQL store?

Example:
- deployed everywhere:
Dicom version: 0.1.1
SQL version: v10

New schema version is rolling out:
Dicom version: 0.1.2
SQL Version: v11

Region A gets the update
SQL replicates v11 to Region B

Region B pod reads v11 from SQL, but the pod is still on version 0.1.1

What happens here? Does region B fallback to v10 or will we get failures if it tries to reference v11?

bcarthic: DICOM always supports n-1 and n+1 versions. In this case region B will be v10 sprocs

The read-only fallback reads the max completed version from the SchemaVersion table without capping to MaximumSupportedVersion This means if Region A migrates to v11 and SQL replicates it to Region B, the Region B pod (still on 0.1.1) will see Current = 11.

This is safe. DICOM compiles both n-1 and n+1 store implementations, so v11 is already available in the 0.1.1 binary. The VersionedCache resolves stored procedures via FirstOrDefault(x => x.Version <= current), which picks the highest compiled-in implementation at or below the current version. Even if a version beyond n+1 range somehow appeared, it would gracefully select the highest available implementation — no exceptions, no failures.

This matches the existing behavior on the primary path, where the [UpsertInstanceSchema]( stored proc can also return versions above [MaximumSupportedVersion] due to an operator precedence issue in its WHERE clause. The system has always relied on VersionedCache as the real safety net, not version capping.

jnlycklama (dismissed):

jnlycklama (approved):

PR #1410, opened 2026-05-27
Add ISchemaWriteGate to allow schema worker to skip writes on read-only SQL secondaries

jnlycklama:
## Description

Fixes a crash-loop bug in services using `SchemaJobWorker` when running against a geo-replication read-only SQL secondary.

**Root cause:** On a geo-replication secondary, Azure SQL routes connections to a read-only replica endpoint. `SchemaJobWorker` calls `UpsertInstanceSchemaInformationAsync` which fails at `connection.Open()` with SQL error 18456 (login routed away from writable endpoint) — *before* the existing `catch (SqlException) when (ex.Number == 3906)` can fire — causing pods to crash-loop on startup.

**Fix:** Introduces `ISchemaWriteGate` — a proactive guard that services can implement to signal whether the current SQL instance is writable. `SchemaJobWorker` resolves the gate from DI each iteration and skips the upsert if `CanWriteAsync()` returns `false`. The existing 3906 catch is retained as a race-condition fallback.

## Changes

- **`ISchemaWriteGate`** — new interface with a default method returning `true`, so services not configured for geo-replication need no changes
- **`DefaultSchemaWriteGate`** — no-op default implementation (always returns `true`), registered via `TryAddSingleton` in `AddSqlServerManagement`
- **`SchemaJobWorker`** — resolves `ISchemaWriteGate` from the DI scope each iteration; skips the upsert if the gate returns `false`

## Testing

- Unit tests added to `SchemaJobWorkerTests`:
  - Gate returns `false` → upsert skipped, schema read from replica
  - Gate returns `true` → normal upsert path
  - Gate returns `true` but 3906 thrown → race-condition fallback handled

## Notes

Companion PR in `dicom-server` wires a `SqlRoleSchemaWriteGate` implementation that delegates to `ISqlRoleDetector.IsPrimaryAsync()` (queries `sys.dm_geo_replication_link_status`, cached 1 min): https://microsofthealth.visualstudio.com/DefaultCollection/Health/_git/dicom-server/pullrequest/62147

### DIFF
--- a/src/Microsoft.Health.SqlServer.UnitTests/Features/Schema/SchemaJobWorkerTests.cs
+++ b/src/Microsoft.Health.SqlServer.UnitTests/Features/Schema/SchemaJobWorkerTests.cs
@@ -30,6 +30,7 @@ public sealed class SchemaJobWorkerTests : IDisposable
     private readonly IProcessTerminator _processTerminator;
     private readonly ILogger<SchemaJobWorker> _logger;
     private readonly ISchemaDataStore _schemaDataStore;
+    private readonly ISchemaWriteGate _writeGate;
     private readonly SchemaJobWorker _worker;
     private int _callCount;
     private readonly CancellationTokenSource _cts = new CancellationTokenSource(1000);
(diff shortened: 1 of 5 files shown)


arunmk-ms (approved):
LGTM

PR #1425, opened 2026-06-15
Expire stale ValueCache<HealthReport> so /health/check surfaces Unhealthy

jnlycklama:
## Summary

Add an expiry to `ValueCache<HealthReport>` so the cached aggregate health report is no longer served indefinitely after the publisher stops publishing fresh reports.

## Background — production scenario

In a Microsoft DICOM Server production deployment, a tenant's SQL database was deleted. After that, the affected pod's `/health/check` endpoint kept returning **200 OK** for **6+ days** even though every internal SQL probe was failing.

What we expected: the per-check `HealthCheckResultCache` would expire, the framework's `HealthCheckService.CheckHealthAsync` would convert the per-check exception into an Unhealthy entry, `HealthCheckCachePublisher.PublishAsync` would write a new Unhealthy report, and the middleware would return 503.

What actually happens: `SqlServerHealthCheck` issues a SQL retry storm that takes 30s+ to fail. The publisher's iteration is cancelled / aborted before the aggregate report is produced, so `PublishAsync` is never called with a fresh report. `ValueCache<HealthReport>` has **no expiry**, so the **6-day-old Healthy report** keeps being served by `CachedHealthCheckMiddleware`.

By contrast, a freshly-started pod returns 503 correctly because `ValueCache<HealthReport>` has not been `Set` yet.

## Fix

1. **`ValueCache<T>`** — add a constructor that accepts a `TimeSpan expiry`. When configured, `GetAsync` returns `null` if the latest `Set` happened longer than the expiry ago. The default constructor preserves the existing infinite-cache behavior, so all current `ValueCache<CustomerKeyHealth>` callers are unaffected.
2. **`HealthReportCachingOptions`** — new options class controlling the expiry of the published `HealthReport` (default 5 minutes; set `Timeout.InfiniteTimeSpan` to disable).
3. **`AddHealthCheckCachePublisher`** — registers `ValueCache<HealthReport>` with the configured expiry; new optional `Action<HealthReportCachingOptions> configureReportCache` parameter is additive (existing callers unchanged).
4. **`CachedHealthCheckMiddleware`** — when `GetAsync` returns `null` (stale), returns `503 Service Unavailable` with a synthetic Unhealthy body. This makes a long-lived pod whose publisher has gone silent behave the same as a brand-new pod whose publisher has not yet succeeded.

## Tests

All passing on net8.0, net9.0, net10.0.

- New `ValueCacheTests` covering fresh / stale / refresh-after-expiry / cancellation / argument validation.
- New `CachedHealthCheckMiddlewareTests` covering fresh Healthy → 200, fresh Unhealthy → 503, and the regression: stale published Healthy report → 503 with `OverallStatus = "Unhealthy"`.
- All existing `Microsoft.Health.Core.UnitTests` (47/47) and `Microsoft.Health.Api.UnitTests` (78/78) still pass.

## Out of scope

The upstream `SqlServerHealthCheck` swallowing only `IsCMKError` / `IsInvalidAccess` and propagating SQL retry-exhaustion exceptions (login-failure 18456, etc.) is a related but separate concern and not part of this PR.

## Compatibility

- `ValueCache<T>()` default constructor unchanged — infinite cache.
- `ValueCache<CustomerKeyHealth>` callers unaffected.
- `AddHealthCheckCachePublisher` adds an optional second parameter; existing call sites keep working.
- Default `HealthReportCachingOptions.Expiry` of 5 minutes is the only behavioral change for existing consumers of `AddHealthCheckCachePublisher`. Set `Expiry = Timeout.InfiniteTimeSpan` to opt out.

### DIFF
--- a/src/Microsoft.Health.Core/Features/Health/HealthReportCachingOptions.cs
+++ b/src/Microsoft.Health.Core/Features/Health/HealthReportCachingOptions.cs
@@ -1,0 +1,29 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Threading;
+
+namespace Microsoft.Health.Core.Features.Health;
+
+/// <summary>
+/// Options that control how the cached <see cref="Microsoft.Extensions.Diagnostics.HealthChecks.HealthReport"/>
+/// published by <c>HealthCheckCachePublisher</c> is treated by readers (e.g. the cached health-check middleware).
+/// </summary>
+public class HealthReportCachingOptions
+{
+    /// <summary>
+    /// Gets or sets the maximum age of a published <c>HealthReport</c> before it is considered stale.
+    /// </summary>
+    /// <remarks>
+    /// When the most recent publish is older than this value, readers receive <see langword="null"/>
+    /// instead of the stale report and the cached health-check middleware responds with
+    /// <c>503 Service Unavailable</c>. This protects against the case where the publisher iteration
+    /// has stopped completing (for example, a health check whose underlying dependency is hanging or
+    /// throwing for longer than the publisher period). Set to <see cref="Timeout.InfiniteTimeSpan"/>
+    /// to disable expiry. Defaults to 5 minutes.
+    /// </remarks>
+    public TimeSpan Expiry { get; set; } = TimeSpan.FromMinutes(5);
+}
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Core/Features/Health/HealthReportCachingOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    /// <remarks>
+    /// When the most recent publish is older than this value, readers receive <see langword="null"/>
+    /// instead of the stale report and the cached health-check middleware responds with
+    /// <c>503 Service Unavailable</c>. This protects against the case where the publisher iteration
+    /// has stopped completing (for example, a health check whose underlying dependency is hanging or
+    /// throwing for longer than the publisher period). Set to <see cref="Timeout.InfiniteTimeSpan"/>
+    /// to disable expiry. Defaults to 5 minutes.
+    /// </remarks>
+    public TimeSpan Expiry { get; set; } = TimeSpan.FromMinutes(5);
```
gunjitchhhatwal, line 37: include the reason for choosing this default value

jnlycklama: Good call - added a <value> block in c119203 explaining the rationale. TL;DR: 5 min is several multiples of the typical publisher Period (30s-1m) so one slow/skipped iteration doesn't flip us to 503, and it exceeds the worst-case single-probe duration (e.g. ~225s SqlClient retry budget for a transient login failure), giving roughly one publish cycle of headroom before the cache is treated as stale.

gunjitchhhatwal (approved):

PR #1433, opened 2026-06-26
Add DeadlockVictim error code to SqlErrorCodes

gunjitchhhatwal:
Added a constant for deadlock victim error code.

Patch


ukachiokoli:
nit, should we have the error codes in numerical order or do we just put the newest one at the top?

CascadeJonathan (approved):

arunmk-ms (approved):

PR #1440, opened 2026-07-09
Catch Aggregate exception

abiisnn:
## Description
When a customer's Key Vault DNS is unresolvable (returns "Name or service not known"), the StorageInitializedHealthCheck returns Unhealthy (503) instead of Degraded (200) with a CMK-related message.

A DNS resolution failure throws HttpRequestException (wrapping SocketException: Name or service not known), which the Azure SDK retry policy wraps in AggregateException. Neither type is in the catch filter, so the exception escapes to CustomerKeyValidationBackgroundService.CheckHealth(), which resets the CMK health cache to healthy by design ("unexpected errors should not be categorized as customer misconfiguration"). This causes StorageInitializedHealthCheck to conclude CMK is fine and return Unhealthy (503) — when the correct response is Degraded (200).

## Impact
AKS readiness probes see 503 → pods get pulled from service
Customer gets full outage instead of a graceful degraded state indicating a CMK misconfiguration
Masks the real issue (Key Vault unreachable) behind a generic "storage not initialized" error

## Example of exception:
exception type: System.AggregateException
exception message: Retry failed after 4 tries. Retry settings can be adjusted in ClientOptions.Retry or by configuring a custom retry policy in ClientOptions.RetryPolicy. (Name or service not known (name-kv.vault.azure.net:443)) (Name or service not known (name-kv.vault.azure.net:443)) (Name or service not known (name-kv.vault.azure.net:443)) (Name or service not known (name-kv.vault.azure.net:443))

## Related issues
Addresses [issue #196329](https://microsofthealth.visualstudio.com/Health/_workitems/edit/196329).
## Testing
Added unit test

Patch|Skip|Feature|Breaking

### DIFF
--- a/src/Microsoft.Health.Encryption/Customer/Health/KeyWrapUnwrapTestProvider.cs
+++ b/src/Microsoft.Health.Encryption/Customer/Health/KeyWrapUnwrapTestProvider.cs
@@ -46,6 +46,18 @@ internal class KeyWrapUnwrapTestProvider : IKeyTestProvider
         }
     }
 
+    internal KeyWrapUnwrapTestProvider(
+        KeyClient keyClient,
+        IOptions<CustomerManagedKeyOptions> cmkOptions,
+        ILogger<KeyWrapUnwrapTestProvider> logger)
+    {
+        EnsureArg.IsNotNull(cmkOptions, nameof(cmkOptions));
+
+        _keyClient = EnsureArg.IsNotNull(keyClient, nameof(keyClient));
+        _customerManagedKeyOptions = EnsureArg.IsNotNull(cmkOptions.Value, nameof(cmkOptions.Value));
+        _logger = EnsureArg.IsNotNull(logger, nameof(logger));
+    }
+
     public async Task<CustomerKeyHealth> AssertHealthAsync(CancellationToken cancellationToken = default)
     {
         if (_keyClient == null)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Encryption/Customer/Health/KeyWrapUnwrapTestProvider.cs
```diff
@@ -71,14 +83,23 @@ public async Task<CustomerKeyHealth> AssertHealthAsync(CancellationToken cancell
...
-
-            return new CustomerKeyHealth
-            {
-                IsHealthy = false,
-                Reason = HealthStatusReason.CustomerManagedKeyAccessLost,
-                Exception = ex,
-            };
+            return HandleKeyAccessFailure(ex);
+        }
+        catch (AggregateException ex) when (ex.Message.Contains("vault.azure.net", StringComparison.OrdinalIgnoreCase))
```
jnlycklama: I would add matching on "Name or service not known" here, so we don't accidentally mark other types of failures as CMK issues. That is a standard error message I have seen across azure services

src/Microsoft.Health.Encryption/Customer/Health/KeyWrapUnwrapTestProvider.cs
```diff
@@ -46,6 +46,18 @@ public KeyWrapUnwrapTestProvider(
         }
     }
 
+    internal KeyWrapUnwrapTestProvider(
```
jnlycklama: Instead of creating a new constructor here, in the tests you could try to set `_customerManagedKeyOptions.KeyVaultUri` with the name of a keyvault that doesn't exist - that will likely get you the same error

src/Microsoft.Health.Encryption/Customer/Health/KeyWrapUnwrapTestProvider.cs
jnlycklama: nit - combine this with the existing `catch` statement since we are handling them in the same way

jnlycklama (approved):

abiisnn:
@microsoft-github-policy-service rerun

PR #1449, opened 2026-07-20
Support VECTOR columns in generated SQL schema models

annagornyitzki:
## Summary

- Generate a schema-only string constant for VECTOR columns in ordinary `CREATE TABLE` statements.
- Update SQL parsing from `TSql150Parser` to `TSql170Parser`.
- Fail generation with file, line, column, and SQL error details when parsing fails.
- Add an end-to-end reproduction for `vector(1536)`.

## Problem

A VECTOR column could generate invalid C# similar to:

    internal readonly Column Embedding = new Column("Embedding", 1536);

`Column` is abstract, so consuming projects fail to compile.

## Fix

For ordinary VECTOR table columns, the generator now emits:

    internal const string Embedding = "Embedding";

The change is limited specifically to VECTOR columns in ordinary `CREATE TABLE` statements. It does not change table-valued parameter behavior or add VECTOR read/write support.

## Validation

- Added the exact `CREATE TABLE dbo.VectorTest` reproduction.
- Focused VECTOR generation test passes.
- Unit-test project builds successfully on `net8.0`, `net9.0`, and `net10.0`.
- Generated test code is written under `obj` and is not committed manually.

## Release Note

FHIR Server currently consumes the v10 package line. This fix may also need to be backported to `release/v10`.


annagornyitzki:
@microsoft-github-policy-service agree company="Microsoft"

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateTableVisitor.cs
```diff
@@ -33,7 +33,7 @@ public override void Visit(CreateTableStatement node)
 
             ClassDeclarationSyntax classDeclarationSyntax =
                 CreateSkeletalClass(className, schemaQualifiedTableName)
-                    .AddMembers(node.Definition.ColumnDefinitions.Select(CreatePropertyForTableColumn).ToArray());
+                    .AddMembers(node.Definition.ColumnDefinitions.Select(CreatePropertyForCreateTableColumn).ToArray());
```
wsugarman, line 36: Why is this separate method necessary? Is that because the `CreatePropertyForTableColumn` is relying on `Enum.TryParse<SqlDbType>` and `SqlDbType.Vector` isn't defined for any Target Framework below .NET 10? If so, I think we can leave this functionality only for .NET 10, unless we need it for an earlier version?

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlModelGenerator.cs
```diff
@@ -66,8 +67,19 @@ private IEnumerable<TSqlFragment> ParseSqlFiles()
...
+                var parser = new TSql170Parser(true);
+                TSqlFragment fragment = parser.Parse(reader, out var errors);
+
+                if (errors.Count > 0)
+                {
+                    string errorDetails = string.Join(
+                        Environment.NewLine,
+                        errors.Select(error => $"({error.Line},{error.Column}) SQL{error.Number}: {error.Message}"));
+
+                    throw new InvalidOperationException($"Failed to parse SQL file '{sqlFile}':{Environment.NewLine}{errorDetails}");
```
wsugarman, line 79: I take it this is a QOL change because `Parse` isn't throw an exception on invalid T-SQL? If so, then this is a good change. Although, I would use `FormatException` instead of `InvalidOperationException`.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlModelGenerator.cs
```diff
@@ -66,8 +67,19 @@ private IEnumerable<TSqlFragment> ParseSqlFiles()
             using (var stream = File.OpenRead(sqlFile))
             using (var reader = new StreamReader(stream))
```
wsugarman, line 68:
```suggestion
            using var stream = File.OpenRead(sqlFile);
            using var reader = new StreamReader(stream);
```

Then you can remove the indent from the lines below

src/Microsoft.Health.SqlServer.UnitTests/Microsoft.Health.SqlServer.UnitTests.csproj
```diff
@@ -2,10 +2,13 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(SupportedFrameworks)</TargetFrameworks>
+    <VectorSchemaModelOutput>$(MSBuildProjectDirectory)\obj\$(Configuration)\$(TargetFramework)\VectorSchema.Generated.cs</VectorSchemaModelOutput>
```
wsugarman, line 5: Should these new tests instead move to `Microsoft.Health.Tools.Sql.Tasks.Tests`?

wsugarman (changes requested):

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/CreateTableVisitor.cs
annagornyitzki: The separate method isn't about `SqlDbType.Vector` enum availability — it's because there's no `VectorColumn` type in the SqlServer schema model. `CreatePropertyForTableColumn` builds a strongly-typed `{SqlDbType}Column` field, but since there's no model type for `vector`, we fall back to emitting the column name as an `internal const string` so the column is still referenceable. That fallback is needed for every target framework that builds vector schema (net8.0/net9.0/net10.0), not just net10.0, so gating it to .NET 10 would break the earlier TFMs.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlModelGenerator.cs
annagornyitzki: Done — now throws `FormatException` instead of `InvalidOperationException`.

src/Microsoft.Health.SqlServer.UnitTests/Microsoft.Health.SqlServer.UnitTests.csproj
annagornyitzki: Done — moved the tests to `Microsoft.Health.Tools.Sql.Tasks.Tests` (scoped to net10.0, since the generator doesn't target net472).

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlModelGenerator.cs
annagornyitzki: Done — switched to `using` declarations and de-indented the block.

PR #1475, opened 2026-09-08
Create First Pass at 1ES Package Pipeline

wsugarman:
This pipeline stubs out the new 1ES pipeline for healthcare shared components so that it may be configured in ADO. It does not clean up the existing pipeline because the validation on the new one has not yet been completed.

This PR also uses apt-get for installing azure functions core tools to avoid NPM.


gunjitchhhatwal (commented):
Could we validate the downloaded `packages-microsoft-prod.deb` against a pinned SHA-256 checksum before installing it? The previous Functions Core Tools installation explicitly verified the downloaded artifact’s digest.

wsugarman:
> Could we validate the downloaded `packages-microsoft-prod.deb` against a pinned SHA-256 checksum before installing it? The previous Functions Core Tools installation explicitly verified the downloaded artifact’s digest.

APT is already verifying the signature and checksum, but I've added verification of the .deb file as well as verifying the local package files after installation.

gunjitchhhatwal (approved):

